PR #11, opened 2025-08-25
docs: Add information about models

lyuzinmaxim:


docs/models.md
```diff
@@ -0,0 +1,73 @@
...
+transform = transforms.Compose([
+    transforms.Resize((256, 256)),
+    transforms.ToTensor(),
+    transforms.Normalize(mean=[0.5, 0.5, 0.5], std=[0.25, 0.25, 0.25])
+])
+```
+## Why Preprocessing Is Required
+
+- **Match training conditions**: Models learn with a specific size (256×256), color channels (RGB), and normalization; inference must match that to avoid accuracy drop.  
+- **Numerical stability**: Normalization (mean/std) keeps activations in ranges the network expects, preventing saturation.
```
lyuzinmaxim: Numerical stability might be an issue during the training, but not during the inference phase. I'd remove this point

docs/models.md
```diff
@@ -0,0 +1,73 @@
...
+    transforms.Resize((256, 256)),
+    transforms.ToTensor(),
+    transforms.Normalize(mean=[0.5, 0.5, 0.5], std=[0.25, 0.25, 0.25])
+])
+```
+## Why Preprocessing Is Required
+
+- **Match training conditions**: Models learn with a specific size (256×256), color channels (RGB), and normalization; inference must match that to avoid accuracy drop.  
+- **Numerical stability**: Normalization (mean/std) keeps activations in ranges the network expects, preventing saturation.  
+- **Consistent layout**: `ToTensor()` converts image data to PyTorch’s channel-first layout `(C,H,W)` and `float32` in `[0,1]`, as required by the model.
```
lyuzinmaxim: Pytorch can support any of the format: (C,H,W), (B, C, H, W), (B, C, W, H) etc. We should see what's the actual input of the network. Since the network from the WeedZSL was based on torchvision models, let's use torchvision documentation to gather this information.

docs/models.md
```diff
@@ -0,0 +1,73 @@
...
+- **Expected tensor shape (PyTorch)**: `(N, C, H, W)` = `(batch, channels, height, width)`.  
+  For these models: **`(1, 3, 256, 256)`**.  
+- **What `unsqueeze(0)` does**: After `transform(image)`, the tensor is `(3, 256, 256)`.  
+  `unsqueeze(0)` adds a batch dimension → `(1, 3, 256, 256)`, which is mandatory because the network processes batches (even of size 1).  
+- **Why batch dim matters**: Enables efficient vectorized execution and lets you pass multiple images at once when needed.  
+
+
+To run inference, use:
+
+```
```
lyuzinmaxim: For this code block and for all others: you can add the word `python` after ` ``` `, and your block will be highlighted with colors which improves the readability. Please try to do that

Issue #15, opened 2025-08-27
Research classification evaluation metrics and choose X metrics to continue with


lyuzinmaxim:
### 🎟️ Research & Add Example Metrics Evaluation

**Description**  
Research classification evaluation metrics and decide on a set of **3 metrics** to continue with (e.g., accuracy, precision, recall, F1-score, macro/micro flavors, ...). Provide a **toy example** demonstrating how to compute them using popular libraries. References:  
- [Google ML Crash Course](https://developers.google.com/machine-learning/crash-course/classification/accuracy-precision-recall)  
- [scikit-learn](https://scikit-learn.org/stable/modules/model_evaluation.html#classification-metrics)
- [torchmetrics](https://lightning.ai/docs/torchmetrics/stable/all-metrics.html) (see "Classification")
- [FiftyOne](https://docs.voxel51.com/api/fiftyone.core.collections.html#fiftyone.core.collections.SampleCollection.evaluate_classifications)

---

#### DoD
- [ ] 3 most important metrics are chosen - please ask @lyuzinmaxim to check that first, and continue working only after approval from his side
- [ ] Create a `/metrics/` directory in the repo  
- [ ] Choose the most easy and straightforward implementation of those 3 metrics
- [ ] Create an unified API for the `ground_truth` and `predictions` if needed
- [ ] Add a script (e.g., `metrics_example.py`) that:  
  - Generates a toy classification dataset (e.g., as a simple list/dict)
  - Computes selected metrics
  - Measures average latency with std
  - Aggregates metrics, latency, and prints a table with it (e.g., pandas dataframe)
- [ ] Document chosen metrics and usage in `docs/metrics.md`  

---

#### 🎯 Acceptance Criteria
- [ ] `/metrics/` contains a runnable toy example script  
- [ ] Metrics are being computed, aggregated and printed as a table
- [ ] Documentation clearly explains which metrics were chosen and why, and how to run the example  

---

**Labels:** `metrics`, `evaluation`, `documentation`  
**Reviewers:** Another student pair

PR #21, opened 2025-09-01
Implement classification metrics (#15)

Sarah5567:
## Description
Implemented evaluation setup based on research of classification metrics.  
Selected **2 metrics** for continuation:
- **Accuracy**
- **F1-score (micro)**

Implementation uses **scikit-learn** for simplicity and reliability.

## Changes
- Added `metrics/` package:
  - `api.py` – unified API (`compute_metrics`)
  - `sklearn_metrics.py` – backend with `scikit-learn`
  - `__init__.py` – exposes API
- Added `metrics_example.py`:
  - Generates toy multiclass dataset
  - Trains simple classifier (Logistic Regression)
  - Computes metrics
  - Prints results as a table with `pandas`


metrics/api.py
```diff
@@ -0,0 +1,17 @@
...
+
+def compute_metrics(
+    y_true: Sequence[int],
+    y_pred: Sequence[int],
+    labels: Optional[Iterable[int]] = None,
+) -> Dict[str, float]:
+    """
+    Unified API: accepts ground truth and predictions (single-label multiclass).
+    Returns: {'accuracy': float, 'f1_micro': float}
+    """
```
Rut-Vahab: Docstring not accurate, should mention all returned metrics

metrics/api.py
```diff
@@ -0,0 +1,17 @@
...
+    return {
+        "accuracy": accuracy_score_skl(y_true, y_pred),
+        "f1_micro": f1_micro_skl(y_true, y_pred, labels=labels),
+    }
```
Rut-Vahab: Only two metrics are returned, should allow extending or passing metrics as parameter

metrics/sklearn_metrics.py
```diff
@@ -0,0 +1,15 @@
+from typing import Iterable, Optional, Sequence
+from sklearn.metrics import accuracy_score, f1_score
+
+
+def accuracy_score_skl(y_true: Sequence[int], y_pred: Sequence[int]) -> float:
+    return float(accuracy_score(y_true, y_pred))
+
```
Rut-Vahab: Wrapper adds little value, could just use sklearn directly

metrics/sklearn_metrics.py
```diff
@@ -0,0 +1,15 @@
...
+
+def f1_micro_skl(
+    y_true: Sequence[int],
+    y_pred: Sequence[int],
+    labels: Optional[Iterable[int]] = None,
+) -> float:
+    # micro averages TP/FP/FN over all classes; for single-label multiclass
+    return float(f1_score(y_true, y_pred, labels=labels, average="micro"))
```
Rut-Vahab: Same here, seems redundant

Rut-Vahab:
Function names with _skl suffix are unnecessary, could be simplified.

Rut-Vahab:
metrics_example.py should be moved under examples/ instead of main code.

chani0343:
## Missing Requirements from Original Task

Good work on the initial implementation! However, several key requirements from the original task (#15) still need to be addressed:

### 1. **Third Metric Required**
The task specifies "3 most important metrics" but only 2 are implemented (Accuracy + F1-micro). Based on our dataset analysis showing **extreme class imbalance** (39 classes, ranging from 1 to 1724 objects per class), we need to add **F1-macro** as the third metric.

F1-macro gives equal weight to all classes regardless of frequency, which is crucial for agricultural classification where rare crop types are still important to identify correctly.

### 2. **Missing Latency Measurement**
The task requires "Measures average latency with std" but there's no timing measurement. Please add:
```python
import time
import statistics
# Measure multiple runs and calculate mean/std of execution times

**Comment by lyuzinmaxim** [comment]
General comment: since you're calculating f1 score directly from sklearn, please also calculate explicitly precision and recall

File: metrics/sklearn_metrics.py (hunk 3 above)
Comment (lyuzinmaxim) [reply metrics/sklearn_metrics.py]: In general it's a nice way to "unify/standardize" the functions, but here since we're using the same library it's redundand

File: metrics/api.py [hunk 5]
```diff
@@ -0,0 +1,17 @@
+from typing import Dict, Iterable, Optional, Sequence
+from metrics.sklearn_metrics import accuracy_score_skl, f1_micro_skl
+
+
```
lyuzinmaxim: Please create a pydantic data model for storing the evaluation results. Just as example:
```
from pydantic import BaseModel, Field

class ClassificationReport(BaseModel):
    accuracy: float = Field(..., , ge=0.0, le=1.0, description="Overall accuracy across all classes")
    f1_score_micro: float <SMTH>
    ....
```

Then the return of `compute_metrics` would be of type `ClassificationReport` which is much more readable:
```
def compute_metrics(
    y_true: Sequence[int],
    y_pred: Sequence[int],
    labels: Optional[Iterable[int]] = None,
) -> ClassificationReport:
   ...
   return ClassificationReport(
        accuracy=accuracy,
        f1_score_macro=f1_macro,
    )

```


What we get if we're using pydantic as a data class:
- auto validation of data type (e.g., `float`), data range (see `ge`, `le`)
- easy conversion to json/dicts/another serialization
- easy to read & transparent
- easy to maintain (extend/inherit/change/...)

lyuzinmaxim:
Please add a documentation explaining the metrics

env/requirements.txt
```diff
@@ -15,6 +15,7 @@ matplotlib==3.8.4
 tqdm==4.66.2
 pillow==10.3.0
 requests==2.32.3
+pydantic>=2.6,<3
```
lyuzinmaxim: please use the fixed version, i.e. `==`

metrics/api.py
```diff
@@ -0,0 +1,73 @@
...
+        accuracy=acc,
+        precision_micro=prec_micro,
+        recall_micro=rec_micro,
+        f1_micro=f1_micro,
+        latency_mean=latency_mean,
+        latency_std=latency_std,
+    )
+
+
+def measure_latencies(
```
lyuzinmaxim: Sorry I am confused about how the latency is being computed. Could you please describe the flow?

metrics/metrics_example.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # 5) Calculate metrics with the unified API (Pydantic model)
+    report: ClassificationReport = compute_metrics(
+        y_true=y_test,
+        y_pred=y_pred,
+        latencies=latencies,
+    )
+
+    # 6) Summary results table (Pandas)
+    df = pd.DataFrame(
```
lyuzinmaxim: I think the conversion to pandas dataframe can be done more easily - please check the docs: https://docs.pydantic.dev/1.10/usage/exporting_models/

lyuzinmaxim: because now we're kinda not using all the features of pydantic and doing things manually.

metrics/api.py
Sarah5567: The function first performs a few warmup calls to avoid measuring initialization overhead.  
Then, for each repetition, it records a start timestamp with `time.perf_counter()`, runs `predict_fn(inputs)`, records the end timestamp, and appends the duration (`t1 - t0`) to a list.

lyuzinmaxim: Could you please explain what's in each "repetition"? How can we use it for the model inference?

Sarah5567: Each repetition is one full inference call of `predict_fn(inputs)` measured separately.  
We record the start time, run the model once, record the end time, and store the duration.  
The list of repetitions lets us compute statistics (e.g., mean and std) to characterize model inference latency.

lyuzinmaxim: Are we gonna have the `predict_fn` in the real code for our classification models?

Sarah5567: Yes. we will pass to `predict_fn` the actual function that performs the prediction for our classification models.

lyuzinmaxim: Ok, I see now, thanks. what does "fn" mean in the function name?

Sarah5567: "fn" is just a shorthand for "function".

metrics/api.py
lyuzinmaxim: Could you please rename this file since it's not reflecting the content.

metrics/api.py
lyuzinmaxim: It's interfering with fn as "false negatives". Please use smth different.

Sarah5567: We changed it to "predict_func"

metrics/api.py
Sarah5567: renamed to `metrics_api.py`, is that fine?

lyuzinmaxim (approved):

Issue #16, opened 2025-08-27
Create a modular evaluation script


lyuzinmaxim:
**Description**  
Based on the chosen evaluation metrics (see Ticket #15, #23), implement a modular script that can evaluate multiple models on the dataset. The script should support both **ResNet** and **MobileNet**, and output the selected metrics in a clear format.  

**Tasks**  
- Implement a Python evaluation script (modular, extendable for models in different format, e.g., ONNX)
- Add an easy-to-use entry point (e.g., `bash run_evaluation.sh` or `python3 run_evaluation.py`).  
- Ensure evaluation runs on the dataset for both ResNet and MobileNet models.  
- Log/print results in a structured format for later analysis.  
- Add documentation on usage and expected outputs.
- Add the eval results to `metrics/evaluation_results.md`, please note that afterwards we would need also to add ONNX in here

**Definition of Done (DoD)**
***FOR BOTH OF THE MODELS!***
- [ ] Script runs evaluation and outputs all chosen metrics for chosen model (ResNet/MobileNet).  
- [ ] Script can be executed with a single command (`bash run_evaluation.sh` or `python3 run_evaluation.py`).  
- [ ] Usage instructions are added to `docs/`.  
- [ ] Evaluation results are shared with the team and added to `metrics/evaluation_results.md`

PR #28, opened 2025-09-03
Add modular evaluation script

Rut-Vahab:
# Create a Modular Evaluation Script (#16)

## Overview
This PR introduces a modular evaluation script for AgStream.  
The script allows evaluation of multiple models on the dataset, currently supporting **ResNet** and **MobileNet**, and outputs metrics in a structured and clear format.

## Changes
- Added `run_evaluation.py`: modular and extendable evaluation script.  
- Added `run_evaluation.sh`: easy command-line entry point.  
- Logging and structured output of metrics for analysis.  
- Documentation on usage and expected outputs added to `docs/`.


dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,221 +1,187 @@
...
-    print("🔬 Creating undersampled classification dataset...")
-    print(f"📁 Data root: {args.data_root}")
-    print(f"🔢 Min objects for inclusion: {args.min_objects}")
-    print(f"🔢 Max objects per class: {args.max_objects}")
-    print("⚠️  STRICT FILTERING: Only classes with 12+ objects will be included")
+    return stats
+
+
+if __name__ == "__main__":
+    create_largest_k_classification_dataset("./dataset", 12, 12)
```
r83575: Please use absolute path

run_evaluation.sh
r83575, line 1: Please move this file to `scripts` folder.

src/inference.py
```diff
@@ -65,12 +65,14 @@ def _forward(self, x: Tensor) -> Tensor:
...
             [
-                transforms.Resize((256, 256)),
+                transforms.Resize((224, 224)),  # ✅ שנה מ-256
                 transforms.ToTensor(),
-                transforms.Normalize(mean=[0.5, 0.5, 0.5], std=[0.25, 0.25, 0.25]),
+                transforms.Normalize(
+                    mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]  # ✅ ImageNet standard  # ✅ ImageNet standard
+                ),
             ]
         )
```
r83575, line 81: Why did you do that?

this a preprocessing from the a model repo
```
transform = transforms.Compose([
    transforms.Resize((256, 256)),
    transforms.RandomRotation(20),
    transforms.RandomHorizontalFlip(),
    transforms.RandomVerticalFlip(),
    transforms.ColorJitter(0.1, 0.1, 0.1, 0.1),
    transforms.ToTensor(),
    transforms.Normalize([0.5]*3, [0.25]*3)
])

```

test_full_dataset.py
r83575, line 1: Are you sure we need this page/file?  
It looks like its functionality is already covered by the new `run_evaluation.py` together with the bash wrapper.  
Might be redundant and could cause duplication/confusion.

dataset_preparation/prepare_crops_dataset.py
chani0343: Fixed! Changed from relative path "./dataset" to absolute path using Path.cwd() / "dataset" which resolves to the current working directory + dataset folder. This ensures the script works consistently across different environments.

run_evaluation.sh
chani0343: Good point! Moving run_evaluation.sh to scripts/ folder for better project organization. Will update in the next commit.

test_full_dataset.py
chani0343: You're right, it's redundant now. I'll remove it.

src/inference.py
chani0343: I changed that.
thanks!!!

lyuzinmaxim: please mention explicitly the normalization constants: `transforms.Normalize([0.5, 0.5, 0.5], [0.25, 0.25, 0.25])`

docs/dataset_preparation.md
lyuzinmaxim, line 1: do we need to change this file?

docs/evaluation_usage.md
```diff
@@ -0,0 +1,110 @@
...
+   Accuracy: 0.2621
+   Precision (Micro): 0.2621
+   Recall (Micro): 0.2621
+   F1-Score (Micro): 0.2621
+   Latency: 0.0595s ± 0.1256s
+
+📊 SUMMARY TABLE:
+================================================================================
+    Model    Accuracy Precision Recall F1-Score     Latency (s)
+   RESNET     0.2423    0.2423  0.2423   0.2423  0.0862 ± 0.0862
```
lyuzinmaxim, line 90: It's quite rare that accuracy, precision, recall and f1 score are the same. Could you please double check the results

docs/evaluation_usage.md
```diff
@@ -0,0 +1,110 @@
...
+5. Folder 80 (unknown): 0.0%
+6. Folder 81 (soft brome): 0.0%
+7. Folder 84 (common wild oat): 0.0%
+8. Folder 86 (rye brome): 0.0%
+9. Folder 87 (knotgrass): 0.0%
+10. Folder 9 (sugar beet four-leaf stage): 0.0%
+
+---
+
+## Requirements
```
lyuzinmaxim: please remove this section

scripts/run_evaluation.sh
lyuzinmaxim, line 1: I would have a python script only instead of this bash script since it's not adding any new functionalities

src/optimal_class_mapping.py
```diff
@@ -0,0 +1,258 @@
...
+        81: 96,
+        40: 255,
+        41: 255,
+        44: 255,
+        48: 255,
+        75: 255,
+        78: 255,
+        82: 255,
+    }
+    return direct_mapping.get(model_class_id, 255)
```
lyuzinmaxim: Why are we defaulting to a 255 class is not found? It shouldn't be the case at all

src/optimal_class_mapping.py
```diff
@@ -0,0 +1,258 @@
...
+    89: "copse-bindweed",
+    91: "common buckwheat",
+    94: "soybean",
+    95: "wild carrot",
+    96: "field mustard",
+    255: "vegetation",
+}
+
+
+def map_prediction(model_class_id):
```
lyuzinmaxim: why we need it?

docs/dataset_preparation.md
chani0343: I don't know when this will change. I'll copy the version from the main now.

src/inference.py
chani0343: ✅ Fixed! The normalization constants are now explicitly written as requested:
transforms.Normalize([0.5, 0.5, 0.5], [0.25, 0.25, 0.25])
Removed the double normalization issue and using the original model preprocessing as specified in the model repo."

docs/evaluation_usage.md
chani0343: Removed the Requirements section as requested.

scripts/run_evaluation.sh
chani0343: Removed the bash script as requested. Users can now use the Python script directly.

docs/evaluation_usage.md
chani0343: I ran it again and it's the same results so here's the explanation that gpt gave me
 Double-checked by re-running evaluation - results confirmed identical (0.2621)!
This is mathematically correct for multiclass micro-averaging:
In single-label multiclass classification, micro-averaging makes all metrics identical by definition:
# Each sample: 1 true class, 1 predicted class
# Therefore: TP + FP = TP + FN = Total samples
Micro Precision = TP/(TP+FP) = TP/Total = Accuracy  
Micro Recall = TP/(TP+FN) = TP/Total = Accuracy
Micro F1 = 2×(P×R)/(P+R) = 2×Acc×Acc/(Acc+Acc) = Accuracy

Our results are mathematically sound! The identical values (0.2621) across runs prove consistent evaluation. This is standard sklearn behavior for multiclass micro-metrics. 🎯

The 26.21% accuracy represents genuine model performance on our 76-class plant classification task.

chani0343: I ran it again and these are the conclusions and results of gpt in the context:

✅ Double-checked by re-running evaluation - results confirmed consistent!

This is mathematically correct for multiclass micro-averaging:

In single-label multiclass classification, micro-averaging makes all metrics identical by definition:

Each sample: 1 true class, 1 predicted class
Therefore: TP + FP = TP + FN = Total samples

Micro Precision = TP/(TP+FP) = TP/Total = Accuracy  
Micro Recall = TP/(TP+FN) = TP/Total = Accuracy
Micro F1 = 2×(P×R)/(P+R) = 2×Acc×Acc/(Acc+Acc) = Accuracy

ResNet18: All metrics = 0.2423 (24.23% accuracy)

MobileNet: All metrics = 0.2621 (26.21% accuracy)

Key findings:

✅ Micro-metrics being identical within each model = Mathematically correct

✅ Different accuracy between models = Expected (MobileNet outperforms ResNet by ~2%)

✅ Consistent results across runs = Proves reliable evaluation

This is standard sklearn behavior for multiclass micro-metrics. 🎯

The results represent genuine model performance on our 76-class plant classification task, with MobileNet achieving superior accuracy over ResNet18."

src/optimal_class_mapping.py
chani0343: ✅ Fixed! Removed the default to 255 completely.

Before: return direct_mapping.get(model_class_id, 255) ❌
Now: raise ValueError(f"Invalid model class ID: {model_class_id}") ✅

No more silent defaults - the function now fails fast on invalid inputs instead of masking errors with 255

src/optimal_class_mapping.py
chani0343: The mapping function is essential because of a mismatch between model and dataset:

Model architecture: 83 classes (IDs: 0, 1, 2, ..., 82)
Dataset structure: 76 classes (IDs: 1, 2, 3, 4, 5, 7, 8, 9, 10, 11, 12, 13, 14, 15, 18, 19, ...)

Without mapping:
Model predicts class 0 → No corresponding folder in dataset ❌
Model predicts class 6 → Should map to dataset folder 7 ❌
With mapping:
Model class 0 → Dataset class 1 ✅
Model class 6 → Dataset class 7 ✅
The model was trained on 83 classes but our evaluation dataset has only 76 classes with non-sequential IDs. This mapping bridges that architectural gap.

docs/dataset_preparation.md
lyuzinmaxim: just do that by merging main into your branch

scripts/run_evaluation.sh
lyuzinmaxim: please add the docs how to use it

src/optimal_class_mapping.py
lyuzinmaxim: cannot we just map by the value (string, e.g. `maize`)?

docs/evaluation_usage.md
lyuzinmaxim: I need to check that

scripts/run_evaluation.sh
chani0343: **Added comprehensive Python script documentation to replace the removed bash script. Updated `docs/evaluation_usage.md` with:**

- **Clear usage examples** for single model and both models
- **Docker usage instructions** 
- **All command-line arguments** explained
- **Removed bash script references** completely

**Users now have complete documentation for using the Python script directly.**"

src/optimal_class_mapping.py
chani0343: ✅ Excellent suggestion! Implemented string-based mapping: model class 0 ("maize") → dataset class 1 ("maize"). Much more robust and maintainable than hard-coded ID mapping.

docs/evaluation_usage.md
chani0343: thanks.

docs/dataset_preparation.md
chani0343: ok

lyuzinmaxim (approved):

PR #29, opened 2025-09-04
Fix with zip

chani0343:


docs/dataset_preparation.md
```diff
@@ -1,6 +1,71 @@
...
+  │   ├── 0/
+  │   ├── 1/
+  │   └── ...
+  ├── classes.json
+  └── objects.json
+  ```
+
+### Quick Alternative
+
+If you only need the ready-to-use classification dataset, extract it directly:
```
lyuzinmaxim, line 54: please add a note that git-lfs should be executed first

lyuzinmaxim (approved):

PR #33, opened 2025-09-07
Revert "Fix with zip"

Rut-Vahab:
Reverts KamaTechOrg/AgStream#29


docs/dataset_preparation.md
```diff
@@ -1,77 +1,6 @@
 # Dataset Preparation
 
-> **Note:** Make sure to install and enable [Git LFS](https://git-lfs.com/) before extracting the dataset.
```
lyuzinmaxim, line 3: You can keep the documentation since some parts would be the same (e.g. instructions)

Issue #31, opened 2025-09-07
Change the second stage of data prep from Random-K to Largest-K


lyuzinmaxim:
We decided to change the undersampling strategy: instead of randomly picking 12 objects per class, we will now choose the **largest 12 objects per class**.

As a reference, see [`undersampling_analysis.ipynb`](https://github.com/user-attachments/files/22194634/undersampling_analysis.ipynb)
  
⚠️ Note: this notebook was partially generated with ChatGPT and might contain issues — please validate the logic carefully before reusing.  

**Tasks**  
- Analyze image size distribution (stage 2), collect statistics
- Update the cropping/undersampling code to use **Largest-K selection**.  
- Update the `.zip` dataset archive with the new crops.  
- Document the size distribution analysis and methodology.  

**DoD**  
- [ ] Statistics about image size distribution (stage 2) are added to `docs/`; documentation reflects the updated undersampling approach.  
- [ ] Code for generating crops is updated to use Largest-K strategy.  
- [ ] Updated `.zip` dataset archive is available and tracked via Git LFS.

PR #34, opened 2025-09-07
Change second stage of data prep from Random-K to Largest-K

chani0343:
✅ Code Changes:
- Updated prepare_crops_dataset.py with Largest-K undersampling
- Objects sorted by bounding box area (largest first)
- Maintains original class IDs without remapping
- Added area calculation: area = (right - left) * (bottom - top)
- Updated statistics file to dataset_stats_largest_k.json

✅ Documentation Updates:
- Added Largest-K strategy section in docs/dataset_preparation.md
- Documented benefits: more representative objects, better training
- Added size distribution analysis documentation
- Updated known issues to include size bias
- Code example showing area calculation and sorting

 Addresses Issue #31 DoD:
- Statistics about image size distribution documented
- Code updated to use Largest-K strategy
- Documentation reflects updated undersampling approach
- Ready for updated .zip dataset archive generation


dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,48 +1,41 @@
...
-    print("\n🔍 Collecting object metadata (fast scan)...")
+    print("\n🔍 Collecting object metadata...")
 
-    # Get valid CSV files
     valid_csv_files = []
     for csv_file in bboxes_dir.glob("*.csv"):
         if csv_file.stat().st_size > 0 and csv_file.stem in image_files:
             valid_csv_files.append((csv_file, image_files[csv_file.stem]))
 
-    print(f"🔍 Found {len(valid_csv_files)} valid CSV-image pairs")
```
lyuzinmaxim, line 36: why removing this?

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,48 +1,41 @@
...
+    # Get all available images
     image_files = {img_file.stem: img_file for img_file in images_dir.glob("*.jpg")}
     print(f"🔍 Found {len(image_files)} image files")
 
-    # Collect only metadata (no image loading!)
+    # Collect metadata
     all_objects_metadata = defaultdict(list)
     processed_files = 0
 
-    print("\n🔍 Collecting object metadata (fast scan)...")
```
lyuzinmaxim, line 28: same here, those are not functional changes, please don't change it

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,48 +1,41 @@
 import pandas as pd
-import random
 import json
 import cv2
 from pathlib import Path
 from collections import defaultdict
 
 
-def create_balanced_dataset_fast(data_root, min_objects=12, max_objects_per_class=12):
+def create_balanced_dataset_with_original_ids(data_root, min_objects=12, max_objects_per_class=12):
```
lyuzinmaxim: I don't like the previous name because it's not reflecting what's happening inside. However, the new name also doesn't reflect it

lyuzinmaxim: See what docsting you've used:
```
Creates dataset with LARGEST objects (Largest-K) instead of random sampling
```

So maybe the function name can be based on it

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -52,11 +45,9 @@ def create_balanced_dataset_fast(data_root, min_objects=12, max_objects_per_clas
                     label_id = int(row["Label_ID"])
                     left, top, right, bottom = int(row["Left"]), int(row["Top"]), int(row["Right"]), int(row["Bottom"])
 
-                    # Basic validation (no image needed)
```
lyuzinmaxim, line 55: same here

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,48 +1,41 @@
...
 
-    print(f"🔍 Found {len(valid_csv_files)} valid CSV-image pairs")
-
     for csv_file, image_path in valid_csv_files:
         processed_files += 1
-
         if processed_files % 500 == 0:
             print(f"📊 Processed {processed_files}/{len(valid_csv_files)} files...")
 
         try:
```
lyuzinmaxim, line 33: this `try-except` block doesn't work since you're just `continue` if exception is occured. Remove it, or add a proper error

dataset_preparation/prepare_crops_dataset.py
lyuzinmaxim: is it possible to split this function into several ones? Because currently a lot of things are being done inside. It's fine to have 1 function like this one which in its turn will have some other functions called inside, e.g.,
```python
def create_balanced_dataset_with_original_ids(...) # <- your function
   all_labels = parse_csv_files() # <- do the csv parsing here
   labels_k_objects = filter_classes_less_than_k_objects() # <- just a filtering of classes
   generate_crops(...) # <- image read/write and other opencv logic goes here
```

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -68,11 +59,10 @@ def create_balanced_dataset_fast(data_root, min_objects=12, max_objects_per_clas
 
                 except Exception:
                     continue
-
         except Exception:
             continue
 
-    # ⚠️ STRICT FILTERING: Only classes with 12+ objects
+    # Filter classes with 12+ objects
```
lyuzinmaxim: same here

docs/dataset_preparation.md
```diff
@@ -1,77 +1,6 @@
 # Dataset Preparation
```
lyuzinmaxim: why do you remove this?

docs/dataset_preparation.md
```diff
@@ -127,48 +56,86 @@ Labels are provided through:
...
 python dataset_preparation/decompress_dataset.py
 
 # Or extract from custom location
 python dataset_preparation/decompress_dataset.py classification_dataset.zip extracted_data/
 ```
 
-**Verification**: Script automatically verifies 912 images and 2 JSON files.
+**Verification:** Script automatically verifies 912 images and 2 JSON files.
+
+---
```
lyuzinmaxim, line 174: same here

docs/dataset_preparation.md
```diff
@@ -127,48 +56,86 @@ Labels are provided through:
...
 
 ### Known Issues
-1. **Background contamination**: Some cropped objects may contain other plants in background (e.g., `classification_dataset/images/0/image_1.png`)
-2. **Multi-plant objects**: Individual crops may have multiple plant types visible
+1. **Background contamination:** Some cropped objects may contain other plants in background
+2. **Multi-plant objects:** Individual crops may have multiple plant types visible
+3. **Size bias:** Largest-K strategy may favor certain growth stages or imaging conditions
```
lyuzinmaxim: why was this added here?

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,48 +1,41 @@
 import pandas as pd
-import random
 import json
 import cv2
```
Sarah5567: I think you should use PIL instead of opencv.

lyuzinmaxim:
General comment: please don't request the changes if there's no need for that. The more lines we have in the PR, the more difficult it's to review it.

dataset_preparation/prepare_crops_dataset.py
chani0343: Thank you for the feedback! I restored the original print statements and did not remove any non-functional lines. All status messages and structure are now preserved as in the original code.

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,221 +1,185 @@
...
     all_objects_metadata = defaultdict(list)
-    processed_files = 0
-
-    print("\n🔍 Collecting object metadata (fast scan)...")
-
-    # Get valid CSV files
     valid_csv_files = []
     for csv_file in bboxes_dir.glob("*.csv"):
         if csv_file.stat().st_size > 0 and csv_file.stem in image_files:
             valid_csv_files.append((csv_file, image_files[csv_file.stem]))
```
lyuzinmaxim: that's not what's stored inside, right?

We store tuples with `csv_file_path` and corresponding `image_path`, so basically it's a pair of values, not `valid_csv_files`.

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,221 +1,185 @@
...
-                        "bbox": [left, top, right, bottom],
-                        "object_id": f"{csv_file.stem}_{idx}",
-                        "csv_file": str(csv_file),
-                    }
+                # Skip invalid bounding boxes (not an error, just data filtering)
+                if left >= right or top >= bottom or left < 0 or top < 0:
+                    continue
 
-                    all_objects_metadata[label_id].append(object_metadata)
+                object_metadata = {
```
lyuzinmaxim: Please create a pydantic class for object metadata

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,221 +1,185 @@
 import pandas as pd
-import random
 import json
 import cv2
 from pathlib import Path
 from collections import defaultdict
 
 
-def create_balanced_dataset_fast(data_root, min_objects=12, max_objects_per_class=12):
+def parse_csv_files(bboxes_dir, image_files):
```
lyuzinmaxim: This function is called `parse_csv_files`. However, here's what is being done inside:
- for given dir of bboxes, and already list of all images, we iterate over csv files
- for valid csv entries we create and add to output the dict called `object_metadata`.

I think you can be more specific in the name of function

dataset_preparation/prepare_crops_dataset.py
lyuzinmaxim: the only difference is channel order (PIL uses RGB, and opencv uses BGR). Please check the outputs

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,221 +1,185 @@
...
 
-                except Exception:
-                    continue
+    return all_objects_metadata
 
-        except Exception:
-            continue
 
-    # ⚠️ STRICT FILTERING: Only classes with 12+ objects
+def filter_classes_with_min_objects(all_objects_metadata, min_objects):
```
lyuzinmaxim: what is `min_objects`? It's not clear from the name of function. Please rename it

dataset_preparation/prepare_crops_dataset.py
lyuzinmaxim, line 1: Please take a look at data models which are being used in this class. For example, answer to the questions:
- What are the inputs of each of the function? (add typings for better readability!)
- What happens if some of the files doesn't have a valid key/value? Instead of adding dicts to list, please consider to use pydantic data models - in this case we will fail fast is some fields are wrong.
- Can we say that "Single responsibility principle" applies to our functions, in other words, are they modular?

dataset_preparation/prepare_crops_dataset.py
```diff
@@ -1,221 +1,174 @@
...
-    # Create class mapping (sequential numbering: 0, 1, 2, ...)
-    class_mapping = {}
-    for i, class_id in enumerate(sorted(eligible_classes.keys())):
-        class_mapping[class_id] = i
-
-    # Now crop and save objects (undersampling to exactly 12)
-    print(f"\n✂️ Creating classification dataset (undersampling to {max_objects_per_class} objects per class)...")
-    print(f"🎯 Processing ONLY {len(eligible_classes)} eligible classes")
 
+def save_largest_k_crops(
```
lyuzinmaxim, line 119: I like the way the code is split now. However, I'd split this function into several functions too:
```python
get_largest_k_object(...) # <- here we output ONLY ObjectMetadata for largest objects
create_crops(...) # <- here we iterate over the outputs of get_largest_k_objects and  use the PIL library to save the crops

File: dataset_preparation/prepare_crops_dataset.py [hunk 16]
```diff
@@ -1,221 +1,174 @@
...
-    stats_file = Path("data") / "classification_dataset" / "dataset_stats.json"
-    mapping_file = Path("data") / "classification_dataset" / "class_mapping.json"
+    return stats, total_objects
 
-    with open(stats_file, "w") as f:
-        json.dump(stats, f, indent=2)
 
-    with open(mapping_file, "w") as f:
-        json.dump(class_mapping, f, indent=2)
+def create_largest_k_classification_dataset(data_root: str, min_objects: int = 12, max_objects_per_class: int = 12):
```
lyuzinmaxim: function has an output, but it's not reflected in typing hints. Remove the return, or adjust the typing

docs/dataset_preparation.md
lyuzinmaxim: Please keep the docs which reflect the 2-stage nature of the dataset

lyuzinmaxim (approved):

Issue #32, opened 2025-09-07
Reconvert updated models to ONNX


lyuzinmaxim:
DoD:
- ONNX models in the repository correspond to the up-to-date pytorch models (after fix)

PR #37, opened 2025-09-08
improve export_to_onnx.py and export the models again(#32)

Sarah5567:
This PR updates the ONNX export workflow so that the exported models **match the latest fixed PyTorch checkpoints**.

### Changes
- Refactored `export_to_onnx.py`:
  - Integrated `ResNetInference` and `MobileNetInference` wrappers.
- Re-exported `resnet18.onnx` and `mobilenet.onnx` to align with the corrected `.pt` models.


Rut-Vahab:
why did you add img ?

scripts/export_to_onnx.py
```diff
@@ -22,52 +25,20 @@ def get_onnx_dir() -> Path:
...
-    Supported: resnet18, mobilenet.
-    """
+def get_model(model_name: str):
     if model_name == "resnet18":
-        from torchvision.models import resnet18
-
-        m = resnet18(weights=None)
-        m.fc = nn.Linear(m.fc.in_features, NUM_CLASSES)
-        return m
+        return ResNetInference(weights_path=get_model_path(model_name))
```
lyuzinmaxim: I was wrong in the example in Slack ([link](https://bootcamps2025.slack.com/archives/C09BW2TN7S8/p1757327072214249?thread_ts=1757325924.672829&cid=C09BW2TN7S8)).

In our base inference class, in the `__init__` method we call `self._initialize_model()` ([link](https://github.com/KamaTechOrg/AgStream/blob/f552b84bcae41b855289bf0f1a976947f719c3ea/src/inference.py#L55))

In derived classes (`ResNetInference` and `MobileNetInference`) we use the directive `super().__init__(...)`, so that means that even though we don't explicitly run `_initialize_model()` in there, it's being executed because it's set up in parent class. ([link](https://github.com/KamaTechOrg/AgStream/blob/f552b84bcae41b855289bf0f1a976947f719c3ea/src/inference.py#L116))

lyuzinmaxim: Please set up all parameters needed for conversion for better readability here:
```python
ResNetInference(device=..., weights_path=get_model_path(model_name), num_classes=...)```

File: scripts/export_to_onnx.py [hunk 2]
```diff
@@ -22,52 +25,20 @@ def get_onnx_dir() -> Path:
...
     elif model_name == "mobilenet":
-        from torchvision.models import mobilenet_v2
-
-        m = mobilenet_v2(weights=None)
-        in_f = m.classifier[-1].in_features
-        m.classifier[-1] = nn.Linear(in_f, NUM_CLASSES)
-        return m
+        return MobileNetInference(weights_path=get_model_path(model_name))
```
lyuzinmaxim: Same here as in https://github.com/KamaTechOrg/AgStream/pull/37/files#r2329915450

scripts/export_to_onnx.py
lyuzinmaxim, line 1: Is it possible to get rid of helper functions `repo_root`, `get_onnx_dir`, and `get_model_path`?

In general, it's nice to have such functions, but these ones are not generalized (e.g., `repo_root` will work only in **the exact same folder layer**, and `get_model_path` just wraps (by hardcoding the path) the model files location). Same for `get_onnx_dir`. In another words, if later we decide to move this script to some other directory, we might break it without noticing.

Please see if just providing the pathes will work for you

scripts/export_to_onnx.py
Sarah5567: You're right, and that's why we don't call `_initialize model()` explicitly. is there something wrong in our code?

scripts/export_to_onnx.py
Sarah5567: we replaced `repo_root` and `get_model_path` functions with a single constant variable:
`MODELS_DIR_PATH = Path(__file__).resolve().parent.parent / "models"`
Is that correct?
Additionally, we’re not sure whether to remove `get_onnx_dir`, because this function also ensures the onnx directory exists (creates it if missing).
What do you think is the right approach?

lyuzinmaxim: I'd say it's better to use the direct relative (to the repository root) path:
```
MODELS_DIR_PATH = "models/"
```

As for the second question, please use some more generic functions to check the folder, for example, smth like this:
```
import shutil
from pathlib import Path


def ensure_clean_directory(directory: Path) -> None:
    shutil.rmtree(directory, ignore_errors=True)
    directory.mkdir(parents=True, exist_ok=True)
```
In this way this function can be reused in many places. I think you can create for that a file called `path_utils.py` in `src/` folder

scripts/export_to_onnx.py
```diff
@@ -1,73 +1,36 @@
...
-    model = try_build_known_arch(model_name)
-    cleaned = {k.replace("module.", ""): v for k, v in state.items()}
-    missing, unexpected = model.load_state_dict(cleaned, strict=False)
-    return model.eval()
+        raise ValueError(f"Unsupported model: {model_name}")
 
 
 def main(model_name: str):
-    model = load_model_from_ckpt(model_name)
+    inference_model = get_model(model_name)
```
lyuzinmaxim: please name it `pytorch_model` - we don't do the inference here

scripts/export_to_onnx.py
```diff
@@ -1,84 +1,44 @@
...
     # Detect parameter dtype (fp16/fp32) and match input accordingly
-    param_dtype = next((p.dtype for p in model.parameters() if p.is_floating_point()), torch.float32)
+    param_dtype = next((p.dtype for p in pytorch_model.model.parameters() if p.is_floating_point()), torch.float32)
 
     # Fixed input size: 256x256, 3 channels, batch=1
     dummy = torch.zeros(1, 3, INPUT_SIZE[0], INPUT_SIZE[1], dtype=param_dtype)
 
     # ONNX path/name
-    onnx_path = (get_onnx_dir() / f"{model_name}.onnx").resolve()
+    onnx_dir = (MODELS_DIR_PATH / "onnx").resolve()
```
lyuzinmaxim: you don't need to use `resolve()` function since MODELS_DIR_PATH is already a Path object, so you can just do `onnx_dir = MODELS_DIR_PATH / "onnx"`

PR #39, opened 2025-09-08
Onnx validation(#25)

### LINKED ISSUE #25: Validate correctness of ONNX model
**Description**  
After exporting the model to ONNX, we need to ensure that it is correctly defined and produces consistent results compared to the original PyTorch model. This step is crucial before integrating the model into evaluation or inference pipelines.  

**Tasks**  
- Check that the ONNX model input and output shapes match the original PyTorch model.  
- Validate numerical accuracy using dummy data (to confirm stability and correctness).  
- Run a forward pass on at least one real test image and compare predictions against the PyTorch version.  

Dummy data examples:
```python
dummy_input = torch.randn(*dims, dtype=torch_dtype)

# Comparison example
assert compare_network_outputs(
            pytorch_out,
            onnx_out,
            rtol=self.rtol,
            atol=self.atol,
        ), "PyTorch model output and corresponding ONNX model output are not the same!"

# Example of a function which checks the outputs
def compare_network_outputs(
    a: Union[np.ndarray, torch.Tensor, List, Tuple],
    b: Union[np.ndarray, torch.Tensor, List, Tuple],
    rtol: float = 1e-03,
    atol: float = 1e-05):
   ...
   return np.allclose(a_numpy, b_numpy, rtol=rtol, atol=atol)
```

**Definition of Done (DoD)** 
> ***FOR BOTH OF THE MODELS!***
- [ ] ONNX model has exactly the same input/output shape as the PyTorch model.  
- [ ] Numerical accuracy is validated with dummy data.  
- [ ] Predictions on one test image are compared and documented (docs/onnx_validation.md).  
- [ ] Results (shapes, accuracy checks) are logged in `docs/onnx_validation.md`.

Sarah5567:
This PR adds validation checks to ensure exported ONNX models are consistent with their original PyTorch versions before integration into inference pipelines.

### Changes
- Verified that ONNX input/output shapes match PyTorch models.
- Added numerical accuracy check with dummy data (`torch.randn`).
- Compared predictions on real test images between PyTorch and ONNX.
- Logged results (shapes, accuracy, prediction match) to `docs/onnx_validation.md`.

Closes #25


scripts/onnx_validation.py
```diff
@@ -0,0 +1,163 @@
...
+    return ort.InferenceSession(onnx_path.as_posix())
+
+_PREPROC = transforms.Compose([
+    transforms.Resize((256, 256)),
+    transforms.ToTensor(),
+    transforms.Normalize(mean=[0.5, 0.5, 0.5], std=[0.25, 0.25, 0.25]),
+])
+
+def _to_onnx_input(img: Image.Image) -> np.ndarray:
+    x = _PREPROC(img).numpy()
```
Rut-Vahab: The preprocessing logic (_PREPROC) duplicates the one in ClassifierInferenceBase; consider reusing a shared function.

scripts/onnx_validation.py
```diff
@@ -0,0 +1,163 @@
...
+    onnx_results = {}
+    img_extensions = {".jpg", ".jpeg", ".png", ".bmp"}
+
+    onnx_input_name = onnx_sess.get_inputs()[0].name
+
+
+    for filename in os.listdir(IMAGES_DIR_PATH):
+        file_ext = os.path.splitext(filename)[1].lower()
+        if file_ext in img_extensions:
+            image_path = os.path.join(IMAGES_DIR_PATH, filename)
```
Rut-Vahab, line 66: In assert_same_predictions, predictions are processed one by one; batching multiple images could improve efficiency.
 In

Rut-Vahab:
Overall code structure is clean and modular.

Unit tests for minimal dummy checkpoints would help ensure robustness.

Minor duplication across ResNet/MobileNet could be reduced by abstracting common logic.

scripts/onnx_validation.py
Sarah5567: I think it's redundant here because only 3 images are processed in this function.

scripts/onnx_validation.py
Sarah5567: The function in `ClassifierInferenceBase` is private. Accessing it is technically possible, but not recommended.

src/inference.py
```diff
@@ -258,4 +258,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
lyuzinmaxim: please add a newline

scripts/onnx_validation.py
```diff
@@ -0,0 +1,163 @@
...
+
+NUM_CLASSES = 83
+INPUT_SIZE = (256, 256)
+
+MODELS_DIR_PATH = Path("models")
+IMAGES_DIR_PATH = Path("data/test_images")
+DEVICE = "cpu"
+INPUT_SHAPE = (1, 3, *INPUT_SIZE)
+
+LOG_PATH = Path("docs/onnx_validation.md")
```
lyuzinmaxim: Docs should be created by people manually. If you need any logs, just save it to the temp folder (don't commit it)

scripts/onnx_validation.py
lyuzinmaxim: Good point. Please consider to create a new class by inheriting from `ClassifierInferenceBase` - `ONNXInference`.

scripts/onnx_validation.py
```diff
@@ -0,0 +1,163 @@
...
+    p = argparse.ArgumentParser("Export PyTorch model to ONNX (fixed 256x256 input)")
+    p.add_argument(
+        "model_name",
+        type=str,
+        choices=["resnet18", "mobilenet"],
+        help="Name of the model to export (must exist in models directory)",
+    )
+    args = p.parse_args()
+
+    main(args.model_name)
```
lyuzinmaxim: please add a newline

scripts/onnx_validation.py
Sarah5567: Maybe we should create a file for the models hierarchy, which will contain all these classes - `ClassifierInferenceBase `, `ONNXInference`, `ResNetInference` etc.

scripts/onnx_validation.py
Sarah5567: Part of this ticket is "_Results (shapes, accuracy checks) are logged in docs/onnx_validation.md._".
So what do you want us to change here?

lyuzinmaxim: Sorry it was not formulated clearly; I meant "described in the doc"

scripts/onnx_validation.py
lyuzinmaxim: You can now create a new file which will just contain the ONNX-related class

Sarah5567: Alright, but it won’t resolve the duplication, right?
Or will this new file contain a class that inherits from `ClassifierInferenceBase`? If it does, the hierarchy and the code  will be a bit unclear, don’t you think?

lyuzinmaxim: This new class (in a new file) can be inherited from `ClassifierInferenceBase`. Or, if you think that some things might be not transparent, I suggest to create a new base class for ONNX

src/onnx_model.py
```diff
@@ -0,0 +1,235 @@
...
+def _to_pil_image(image: Union[np.ndarray, Image.Image]) -> Image.Image:
+    if isinstance(image, np.ndarray):
+        return Image.fromarray(image).convert("RGB")
+    elif isinstance(image, Image.Image):
+        return image.convert("RGB")
+    else:
+        raise TypeError(f"Unsupported image type: {type(image)}")
+
+
+class ClassifierInferenceBase(ABC):
```
lyuzinmaxim: we don't need to specify it here - let's just inherit from existing one

src/onnx_model.py
```diff
@@ -0,0 +1,235 @@
...
+            device=device,
+            weights_path=weights_path,
+            class_mapping=class_mapping,
+            topk=topk,
+            transform=transform,
+            **kwargs,
+        )
+
+
+class MobileNetOnnxInference(OnnxClassifierInferenceBase):
```
lyuzinmaxim: does it make sense to differentiate between ResNetONNX and MobileNetONNX? Because the interface seems to be the same

Sarah5567: So what do you think is the best and clearest architecture? because we understood that that's what you ment [here](https://github.com/KamaTechOrg/AgStream/pull/39#discussion_r2333409965) to do.

src/inference.py
```diff
@@ -259,3 +259,4 @@ def main():
 
 if __name__ == "__main__":
     main()
+
```
lyuzinmaxim: Please remove trailing whitespaces - I think running black locally will resolve it

src/onnx_model.py
```diff
@@ -0,0 +1,87 @@
...
+                "EXTENDED": ort.GraphOptimizationLevel.ORT_ENABLE_EXTENDED,
+                "ALL": ort.GraphOptimizationLevel.ORT_ENABLE_ALL,
+            }
+            so.graph_optimization_level = level_map.get(self._opt_level.upper(), ort.GraphOptimizationLevel.ORT_ENABLE_ALL)
+
+        self.session = ort.InferenceSession(self.weights_path.as_posix(), sess_options=so, providers=self._providers)
+        self.input_name = self.session.get_inputs()[0].name
+        self.output_name = self.session.get_outputs()[0].name
+
+    @torch.inference_mode()
```
lyuzinmaxim: I think this decorator isn't needed since it's a ONNX model inference, so no gradients are accumulated

scripts/onnx_validation.py
```diff
@@ -0,0 +1,112 @@
...
+            num_classes=NUM_CLASSES,
+            strict=True,
+        )
+    else:
+        raise ValueError(f"Unsupported model: {model_name}")
+    onnx_model = OnnxClassifierInferenceBase(
+            device=DEVICE,
+            weights_path=MODELS_DIR_PATH / "onnx" / f"{model_name}.onnx",
+            class_mapping=ID_TO_NAME,
+            topk=5,
```
lyuzinmaxim: We don't use the topk now, so please use `topk=1`

lyuzinmaxim (approved):

PR #50, opened 2025-09-15
Prepare nvinfer element with our models

### LINKED ISSUE #49: Prepare nvinfer element with our models
**Description**  
Set up a toy DeepStream pipeline to test ONNX model inference using the `nvinfer` element. The goal is to validate that the model can run inside DeepStream in a minimal working pipeline (fakesink, fakesrc).

References:
- DeepStream docs (**most important**)
  - [general](https://docs.nvidia.com/metropolis/deepstream/6.3/dev-guide/text/DS_Quickstart.html)
  - [gst-nvinfer](https://docs.nvidia.com/metropolis/deepstream/6.3/dev-guide/text/DS_plugin_gst-nvinfer.html)
- [DeepStream python sample apps](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/tree/master?tab=readme-ov-file#sample-applications)
- [DeepStream extra libraries](https://github.com/NVIDIA-AI-IOT/deepstream_libraries) (only if needed)
- Gstreamer base elements
  - [fakesrc](https://gstreamer.freedesktop.org/documentation/coreelements/fakesrc.html)
  - [fakesink](https://gstreamer.freedesktop.org/documentation/coreelements/fakesink.html?gi-language=c#fakesink-page)

**Tasks**
- nvinfer config files (e.g. `mobilenet.txt` or `resnet.txt`) created
  - pre-processing is the same as the models use!
- Configure a simple GStreamer pipeline with fake input → `nvinfer` → fake output.  
- Load the ONNX model in `nvinfer`.  
- Confirm that the pipeline runs end-to-end without errors.  

**DoD**  
- [ ] Minimal DeepStream pipeline with `nvinfer` successfully runs ONNX model.  
- [ ] Logs confirm inference is being executed.  
- [ ] Script to run in a new folder, and pipeline configuration (e.g. model configs) are committed

chani0343:
##  Prepare nvinfer element with our models

This PR implements a minimal DeepStream pipeline to validate ONNX model inference using the `nvinfer` element.

###  What's included:
- **Working DeepStream pipeline**: `videotestsrc → nvvideoconvert → nvstreammux → nvinfer → fakesink`
- **ONNX model configs**: Ready-to-use configurations for MobileNet and ResNet18
- **Test script**: `test_onnx_pipeline.py` for automated validation
- **TensorRT integration**: Successfully builds and runs optimized engines

###  Key achievements:
- Both models load and run inference successfully
- TensorRT engines are generated and cached for performance
- Pipeline completes end-to-end without errors
- Logs confirm inference execution

###  Usage:
```bash
python3 src/test_onnx_pipeline.py --config src/configs/mobilenet_onnx.txt
python3 src/test_onnx_pipeline.py --config src/configs/resnet18_onnx.txt
```

closes #49


src/configs/mobilenet_onnx.txt
```diff
@@ -0,0 +1,18 @@
...
+network-mode=0
+num-detected-classes=83
+gie-unique-id=1
+network-type=1
+classifier-threshold=0.5
+process-mode=1
+model-color-format=0
+maintain-aspect-ratio=1
+symmetric-padding=1
+offsets=127.5;127.5;127.5
```
r83575: Duplicate `offsets` entry – remove one to avoid config parsing issues.

.nv/ComputeCache/0/a/e8cab30dc9453a
r83575, line 1: `.engine` files and `.nv/ComputeCache/` should not be committed.  
Please add them to `.gitignore` to avoid pushing generated or cache files.

src/test_onnx_pipeline.py
```diff
@@ -0,0 +1,70 @@
...
+gi.require_version('Gst', '1.0')
+from gi.repository import Gst, GLib
+import argparse
+
+def main(config_file):
+    Gst.init(None)
+    
+    pipeline = Gst.Pipeline()
+    
+    source = Gst.ElementFactory.make("videotestsrc", "source")
```
r83575, line 20: The pipeline uses `videotestsrc`, but the task specifically requires `fakesrc`.  
Please replace `videotestsrc` with `fakesrc` to match the task requirements.

src/configs/mobilenet_onnx.txt
```diff
@@ -0,0 +1,18 @@
+[property]
+gpu-id=0
+net-scale-factor=0.01568627450980392
+offsets=127.5;127.5;127.5
+onnx-file=/workspace/models/onnx/mobilenet.onnx
+model-engine-file=/workspace/models/onnx/mobilenet.engine
+labelfile-path=/workspace/src/configs/labels.txt
```
r83575: Absolute paths are used in the config files (e.g. `/workspace/models/...`).  
It's better to use relative paths instead, so the code is easier to run in different environments (like dev, test, or prod setups).

r83575 (commented):
The pipeline mostly meets the task goals!

src/configs/mobilenet_onnx.txt
```diff
@@ -0,0 +1,17 @@
+[property]
+gpu-id=0
+net-scale-factor=0.01568627450980392
+offsets=127.5;127.5;127.5
+onnx-file=models/onnx/mobilenet.onnx
+model-engine-file=models/onnx/mobilenet.engine
```
lyuzinmaxim: Let's create a separate dir for the deepstream engines, and add it to the .gitignore for now

src/configs/mobilenet_onnx.txt
```diff
@@ -0,0 +1,17 @@
+[property]
+gpu-id=0
+net-scale-factor=0.01568627450980392
+offsets=127.5;127.5;127.5
```
lyuzinmaxim: These are correct numbers, great work!

Could you please create a documentation on how to calculate them, and how are they related to the preprocessing which we have in [pytorch inference class](https://github.com/KamaTechOrg/AgStream/blob/06e3bf86431579b15bd3e067d3eb50cff2af1027/src/inference.py#L77). In this doc, please refer to the nvidia deepstream nvinfer documentation and maybe to the source code of the nvinfer plugin

src/configs/mobilenet_onnx.txt
lyuzinmaxim, line 1: For the whole file: it would be super useful for everyone if you could add simple comments which will explain "what is this parameter" to the ones which are not clear from the first look:
```
gpu-id=0 # <- no comment needed since it's clear
onnx-file=... # <- same, pretty clear
...
net-scale-factor=... # <- It's confusing so the comment is needed
model-color-format=... # <- It would be nice to give a small description and available options, e.g. "# 0: RGB 1: BGR 2: GRAY"
```

src/configs/resnet18_onnx.txt
lyuzinmaxim, line 1: same as https://github.com/KamaTechOrg/AgStream/pull/50/files#r2355263908

src/configs/mobilenet_onnx.txt
Sarah5567: Are you sure it’s possible to do that? We tried it, but the code didn’t run properly. It looks like adding comments in such a file makes it unreadable.

src/configs/resnet18_onnx.txt
chani0343: Additionally, we configured the engine file path to point directly to /workspace/deepstream_engines/resnet18.engine. This way:
First run: DeepStream creates the engine with auto-generated name, our script copies it to the clean name and deletes the original
Subsequent runs: DeepStream finds the existing clean-named engine file and uses it directly, skipping rebuild
This gives us predictable file names while maintaining DeepStream's caching.

src/configs/mobilenet_onnx.txt
lyuzinmaxim: I was wrong, the comment should be at its own line, not after the actual string, e.g.
```
# 0: FP32 1: INT8 2: FP16
network-mode=2
```

lyuzinmaxim: Please consider rename this and resnet file to not have `onnx` in the filename

lyuzinmaxim: Also please add comments (as suggested previously) to the config files directly instead of readme.

src/configs/labels.txt
lyuzinmaxim, line 1: don't you want to add here the real class names?

src/configs/mobilenet_onnx.txt
Sarah5567: do you mean [this file](https://github.com/KamaTechOrg/AgStream/blob/fb765094e167126c4b320c95631e387e5266926d/docs/DeepStream_preprocessing.md)? Should we remove it and move its content into the config files?

lyuzinmaxim (approved):

PR #58, opened 2025-09-29
implement DeepStream pipeline end-to-end

### LINKED ISSUE #56: Implement the DeepStream pipeline and test end-to-end
**Description**  
We want to implement an end-to-end DeepStream pipeline that takes an RTSP input stream, performs inference, overlays results, and saves processed frames to disk. This will serve as the first complete working version of our inference pipeline.  

**Pipeline:**  
[RTSP input] -> [video parsing/decoding] -> [conversion] -> [model inference] -> [OSD] -> [save the frames as images]

Links:
- [deepstream OSD plugin](https://docs.nvidia.com/metropolis/deepstream/6.4/dev-guide/text/DS_plugin_gst-nvdsosd.html)

**Tasks**  
- Build a DeepStream pipeline with the listed elements.  
- Integrate the OSD plugin to overlay timestamp, frame number, and model predictions.  
- Save frames as images on disk for manual inspection.  
- Add cleanup logic to manually remove saved frames to avoid disk bloat.  
- Document usage and expected output.  

**DoD**  
- [x] Pipeline runs end-to-end from RTSP input to saved images.  
- [x] Output images contain OSD overlays (timestamp, frame info, predictions).  
- [x] Disk cleanup command is created.

Sarah5567:
Implement an end-to-end **DeepStream pipeline**:  
This provides the first working version of the inference pipeline with documentation and cleanup utilities.  

Our pipline:
- Reads RTSP input
- Decodes and converts video
- Runs AI inference
- Overlays results (timestamp, frame number, predictions)
- Saves processed frames as images
 
closes #56


src/deepstream_pipeline.py
```diff
@@ -0,0 +1,64 @@
...
+    
+    pipeline_str = (
+        f"rtspsrc location={RTSP_URL} latency=200 ! "
+        "rtph264depay ! h264parse ! decodebin ! videoconvert ! "
+        "nvvideoconvert ! "
+        "m.sink_0 nvstreammux name=m batch-size=1 width=640 height=480 ! "
+        f"nvinfer config-file-path={CONFIG_FILE} ! "
+        "nvdsosd display-clock=1 display-text=1 clock-font-size=20 ! "
+        "nvvideoconvert ! "
+        "jpegenc ! "
```
Rut-Vahab: Consider wrapping Gst.parse_launch with try/except to catch errors when creating the pipeline.

src/deepstream_pipeline.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+
+import gi
+gi.require_version('Gst', '1.0')
+from gi.repository import Gst, GLib
+import os
+
+RTSP_URL = "rtsp://127.0.0.1:8554/test"
+OUTPUT_DIR = "/workspace/output/frames/"
+CONFIG_FILE = "/workspace/src/configs/resnet18.txt"
```
Rut-Vahab, line 14: Consider move RTSP_URL, OUTPUT_DIR, CONFIG_FILE into environment variables or a config file for easier maintenance.

src/deepstream_pipeline.py
```diff
@@ -0,0 +1,64 @@
...
+
+def main():
+    Gst.init(None)
+    os.makedirs(OUTPUT_DIR, exist_ok=True)
```
Rut-Vahab: Consider split main() into smaller functions like build_pipeline(), handle_messages(), etc., to improve readability and maintainability. (lines 18–64)

src/deepstream_pipeline.py
```diff
@@ -0,0 +1,64 @@
...
+            loop.quit()
+    
+    bus.connect("message", on_message)
+    pipeline.set_state(Gst.State.PLAYING)
+    
+    print(f"✅ Pipeline started - saving frames to {OUTPUT_DIR}")
+    print("📸 Frames will include OSD overlays (timestamp, predictions)")
+    print("Press Ctrl+C to stop")
+    
+    loop = GLib.MainLoop()
```
Rut-Vahab: Add a SIGTERM handler in addition to Ctrl+C for graceful shutdown, useful when running in scripts

src/deepstream_pipeline.py
lyuzinmaxim: It's a nice idea, but I'd keep it as is for now because of the simplicity

src/deepstream_pipeline.py
lyuzinmaxim: I wouldn't mix the streamer and OS logic together - please do all OS operations before/after gstreamer calls

docs/deepstream_pipeline.md
```diff
@@ -0,0 +1,40 @@
+# DeepStream End-to-End Pipeline
+
+## Overview
+Complete DeepStream pipeline: RTSP → Decode → Convert → AI Model → OSD → Save Images
```
lyuzinmaxim: I personally think that "AI model" is more a term for e.g. business, I'd be more specific, for example, "classification model" or "nvinfer"

docs/deepstream_pipeline.md
```diff
@@ -0,0 +1,40 @@
...
+
+### 4. Cleanup Frames
+```bash
+bash ./scripts/cleanup_frames.sh
+```
+
+## Output
+
+Frames: `output/frames/frame_00001.jpg`, `frame_00002.jpg`, etc.  
+Contains: timestamp, frame info, AI predictions
```
lyuzinmaxim: same here as https://github.com/KamaTechOrg/AgStream/pull/58/files#r2435012822

src/deepstream_pipeline.py
```diff
@@ -0,0 +1,74 @@
...
+
+RTSP_URL = "rtsp://127.0.0.1:8554/test"
+OUTPUT_DIR = "/workspace/output/frames/"
+CONFIG_FILE = "/workspace/src/configs/resnet18.txt"
+
+
+def build_pipeline():
+    """Build and return a GStreamer pipeline."""
+    pipeline_str = (
+        f"rtspsrc location={RTSP_URL} latency=200 ! "
```
lyuzinmaxim: There're several ways on how to declare the pipeline. The first one is to do as you did - using the CLI and passing as a string to the deepstream. The second way is to link elements in code, e.g. as [here](https://github.com/KamaTechOrg/AgStream/blob/e82b267aaa45e848f0067d9f1ecf1e405968648c/src/test_onnx_pipeline_real.py#L50C9-L50C17). Also you can check the [deepstream demos](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/master/apps/deepstream-test1/deepstream_test_1.py).

For this pipeline, I'd prefer the second one, because:
- It allows you to use custom modifications, such as callbacks, probes etc
- We define all connections between element explicitly, and do not rely on deepstream CLI hidden functionalities (such as ghost pad requesting as in `m.sink_0 nvstreammux name=m batch-size=1 width=640 height=480`)

I'd ask you to change the code to use the second approach

Sarah5567: Are you sure it’s needed here, where we don’t use a pad probe?
Should we do it this way because that’s how we’ll implement it in the final pipeline?

lyuzinmaxim: we will use this script as a base for the future development

PR #63, opened 2025-10-16
Access the raw frames

### LINKED ISSUE #57: Access the raw frames in a deepstream pipeline
**Description**  
We need to add the ability to access raw frames inside a DeepStream pipeline. This allows us to apply custom operations outside of the standard DeepStream plugins (as a part of callback functions). Try to use a pipeline as simple as possible (e.g. without a nvinfer inside, try to remove the "heavy" elements from it), but having important elements (nvvideoconvert). 

As a reference, please use the way of creating the deepstream pipeline as in `test_onnx_pipeline.py` (element factory, adding to the bin, etc).

Links:
- https://github.com/NVIDIA-AI-IOT/deepstream_python_apps

**Tasks**  
- Identify the correct probe/hook/callback point in the pipeline (e.g., pad probe after `nvvidconv`).  
- Implement code to extract raw frames from the pipeline.  
- Demonstrate a simple manipulation on the part of the frame (e.g., flip, brightness change, or drawing text).  
- Document the mechanism clearly for future extensions.  

**DoD**  
- [ ] Raw frames can be accessed from the DeepStream pipeline.  
- [ ] At least one manipulation is applied to confirm access works.
- [ ] The mechanism is documented in `docs/access_raw_frames.md`.

Sarah5567:
**Description**
This PR adds the ability to access and modify raw frames inside a DeepStream pipeline.  
A pad probe is attached after the `nvvideoconvert` (RGBA conversion) element, allowing direct access to GPU buffers via `pyds.get_nvds_buf_surface`.

**Key Features**
- Extracts raw frames from the pipeline (in RGBA format).
- Demonstrates custom manipulation using NumPy.
- Preserves GPU memory integrity using `np.copyto` to write back modified data.
- Designed as a minimal DeepStream pipeline (no `nvinfer` or OSD).
- Compatible with RTSP input and `multifilesink` output.

**Files Added**
- `src/access_raw_frames.py` - main implementation.
- `env/arm/pyds-1.1.10-py3-none-linux_aarch64.whl` - for installing the `pyds` library, which is required for the code.
- `docs/access_raw_frames.md` - documentation describing mechanism and usage.

closes #57


src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+    try:
+        loop.run()
+    except KeyboardInterrupt:
+        print("\nStopping...")
+    finally:
+        pipeline.set_state(Gst.State.NULL)
+        print("Pipeline stopped")
+
+if __name__ == "__main__":
+    main()
```
Rut-Vahab: Please add a new line.

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+RTSP_URL = "rtsp://127.0.0.1:8554/test"
+OUTPUT_DIR = "/workspace/outputs/frames/"
+
+def buffer_probe(pad, info):
+    buf = info.get_buffer()
+    if not buf:
+        return Gst.PadProbeReturn.OK
+
+    # Retrieve NvBufSurface from GPU memory
+    surface = pyds.get_nvds_buf_surface(hash(buf), 0)
```
Rut-Vahab, line 24: Avoid using hash(buf) to access the surface. It’s unreliable across DeepStream versions and may cause invalid memory access.

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Copy to CPU (RGBA format)
+    frame = np.array(surface, copy=True, order='C')
```
Rut-Vahab, line 27: This may not return a valid NumPy array — the surface might be a pointer, not the raw frame data.

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # ---- NumPy-only modification ----
+    # Example: draw a red rectangle at the top of the image
+    h, w, _ = frame.shape
+    frame[0:20, 0:w] = [255, 0, 0, 255]  # red bar, 20 pixels height
+    # Another example: darken the whole frame by 30%
+    # frame = (frame * 0.7).astype(np.uint8)
+
+    # Copy back to GPU
+    np.copyto(np.array(surface, copy=False, order='C'), frame)
```
Rut-Vahab, line 34: Copying back to GPU may fail if the surface isn’t in matching format (RGBA).

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+    elems["depay"].link(elems["parse"])
+    elems["parse"].link(elems["decode"])
+    elems["decode"].link(elems["convert1"])
+    elems["convert1"].link(elems["nvvideoconvert1"])
+
+    # Now link nvvideoconvert1 → capsfilter → streammux
+    elems["nvvideoconvert1"].link(elems["capsfilter"])
+    sinkpad = elems["streammux"].get_request_pad("sink_0")
+    srcpad = elems["capsfilter"].get_static_pad("src")
+    srcpad.link(sinkpad)
```
Rut-Vahab, line 105: streammux should receive raw video, not post-RGBA capsfilter. This link order may cause performance loss.

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+    elems["streammux"].link(elems["nvvideoconvert2"])
+    elems["nvvideoconvert2"].link(elems["jpegenc"])
+    elems["jpegenc"].link(elems["sink"])
+
+
+def connect_dynamic_links(elems):
+    def on_pad_added(src, new_pad):
+        sink_pad = elems["depay"].get_static_pad("sink")
+        if not sink_pad.is_linked():
+            new_pad.link(sink_pad)
```
Rut-Vahab, line 116: Missing pad type check — if RTSP includes audio, it’ll try linking it too and fail.

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+    bus = pipeline.get_bus()
+    bus.add_signal_watch()
+    loop = GLib.MainLoop()
+    bus.connect("message", on_message, loop)
+    return loop
+
+
+def main():
+    Gst.init(None)
+    os.makedirs(OUTPUT_DIR, exist_ok=True)
```
Rut-Vahab, line 143: Create the output directory before building the pipeline, otherwise multifilesink may not find it.

src/access_raw_frames_pipeline.py
Sarah5567: I think that in this case, RGBA conversion before `nvstreammux` is intentional, since the pad probe requires RGBA frames for CPU-side access and NumPy processing. The small performance cost is acceptable for this workflow.

src/access_raw_frames_pipeline.py
lyuzinmaxim: deepstream sample apps show that it's ok to use it:
https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-imagedata-multistream/deepstream_imagedata-multistream.py#L121

src/access_raw_frames_pipeline.py
lyuzinmaxim: I don't know how that works internally, but since we were able to see the output frames manipulated, I suppose that worked properly

src/access_raw_frames_pipeline.py
lyuzinmaxim: In general, yes - but here we don't care about audio stream in video

PR #65, opened 2025-10-19
Access the raw frames

lyuzinmaxim:


src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,167 @@
...
+            new_pad.link(sink_pad)
+
+    elems["rtspsrc"].set_property("location", RTSP_URL)
+    elems["rtspsrc"].set_property("latency", 200)
+    elems["rtspsrc"].connect("pad-added", on_pad_added)
+    elems["decode"].connect("pad-added", on_decode_pad_added)
+
+
+
+def attach_buffer_probe(elems, buffer_probe):
```
lyuzinmaxim: I don't think this function is needed, for me it's confusing since it expects capsfilter to be present in the `elems` list. Let's just not have it as a function, and move the logic to `link_pipeline_elements` (I am not sure logically but seems like it's the place for it)

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,167 @@
...
+    if msg.type == Gst.MessageType.ERROR:
+        err, debug = msg.parse_error()
+        print(f"ERROR: {err}, debug: {debug}")
+        loop.quit()
+    elif msg.type == Gst.MessageType.EOS:
+        print("End of stream")
+        loop.quit()
+
+
+def create_elements():
```
lyuzinmaxim: I see that your idea was to split the logic, but for me it's easier to follow if we create elements, set up their properties and  link them together as one function. Also that function could include the probe attachment

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,167 @@
...
+    sink.set_property("location", os.path.join(OUTPUT_DIR, "frame_%05d.jpg"))
+
+    elements = [rtspsrc, depay, parse, decode, convert1, nvvideoconvert1,
+                capsfilter, streammux, nvvideoconvert2, jpegenc, sink]
+
+    if not all(elements):
+        print("ERROR: Failed to create one or more GStreamer elements", file=sys.stderr)
+        sys.exit(1)
+
+    return {
```
lyuzinmaxim: as I mentioned [here](https://github.com/KamaTechOrg/AgStream/pull/65/files#r2442913582), we wouldn't need to define our own way of transferring the elements between functions

src/access_raw_frames_pipeline.py
```diff
@@ -0,0 +1,167 @@
...
+from gi.repository import Gst, GLib
+import os
+import sys
+import numpy as np
+import pyds
+
+RTSP_URL = "rtsp://127.0.0.1:8554/test"
+OUTPUT_DIR = "/workspace/output/frames/"
+
+def buffer_probe(pad, info):
```
lyuzinmaxim: I like the way this function is split from all others, but the name `buffer_probe` is not describing what's going on inside:
- requesting the buffer
- get surface from buffer
- GPU surface -> CPU 
- CPU manipulations
- CPU surface -> GPU

lyuzinmaxim: please consider to rename it

src/access_raw_frames_pipeline.py
lyuzinmaxim, line 1: General comment: Please refactor based on all other comments, and afterwards please add typings. Also please run linting at the end

PR #72, opened 2025-10-20
add MongoDB deployment and usage documentation

### LINKED ISSUE #68: Deploy a simple database and ensure it's working
Based on the ticket #61, we need to deploy a database (a very simple set up), create scripts for up/down/cleanup/..., and create  usage documentation. 

DoD:
- database is up and running
- database entrypoint is shared, and it's reachable from the jetson

Rut-Vahab:
# PR Summary: MongoDB Deployment

- Deployed a simple MongoDB container for AgStream.
- Created scripts: `up.sh`, `down.sh`, `cleanup.sh`.
- Database is up, reachable from Jetson, and ready for use.

closes #68


database/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: "3.8"
+
+services:
+  mongodb:
+    image: mongo:6.0
+    container_name: agstream_mongo
+    restart: unless-stopped
+    ports:
+      - "27017:27017"
```
Sarah5567, line 9: I think it might cause an error if two people try to connect at the same time, because the port is the same for everyone.

database/scripts/down.sh
Sarah5567, line 1: consider combining all the scripts into one file and using flags to determine which operation to perform.

Rut-Vahab: That was the mission requirement to create scripts.
https://github.com/KamaTechOrg/AgStream/issues/68#issue-3531361111:~:text=set%20up)%2C%20create-,scripts,-for%20up/down

database/docker-compose.yml
Rut-Vahab: The comment isn’t correct in this case. We have a single MongoDB instance running on the Jetson, and multiple users or containers can connect to it at the same time without any problem. MongoDB supports multiple simultaneous connections, so sharing the same port does not cause errors.

Sarah5567: this setup works only if a single MongoDB instance runs on the Jetson.
if multiple users run their own containers, they must use different host ports
(e.g. 27018:27017, 27019:27017) to avoid port binding conflicts.

r83575: You're correct that multiple MongoDB instances on the same host would require different host ports.
However, in our setup we have only a single MongoDB instance running on the Jetson, and multiple users or containers simply connect to that single instance.
We validated this by testing concurrent connections from two separate containers using the same port (27017), and both were able to connect and operate simultaneously without any issue.

docs/mongodb_setup.md
```diff
@@ -0,0 +1,34 @@
...
+```bash
+cd ~/AgStream/database/scripts
+./up.sh      # Start MongoDB
+./down.sh    # Stop MongoDB
+./cleanup.sh # Cleanup all data and containers
+```
+
+## Connection
+
+Use the following connection string from Jetson or host:
```
lyuzinmaxim: jetson = host, I think you meant `from Jetson or container`

docs/mongodb_setup.md
```diff
@@ -0,0 +1,34 @@
...
+
+```bash
+mongosh mongodb://192.168.50.11:27017/agstream
+```
+
+## Notes
+
+* Run scripts from host, not inside another container.
+* `docker compose` must be installed on the host.
```
lyuzinmaxim: don't we have docker compose installed on host (jetson)?

env/arm/Dockerfile
```diff
@@ -28,8 +28,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     gstreamer1.0-plugins-ugly \
     gstreamer1.0-libav \
     libcairo2-dev pkg-config \
+    curl gnupg \
  && rm -rf /var/lib/apt/lists/*
 
+RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - && \
```
lyuzinmaxim: Can we have a separate container for the database? Because the database shouldn't be connected to our ARM container

docs/mongodb_setup.md
lyuzinmaxim, line 1: can you please add information about how to push data to it, and how to check if the "write" is successful?

Also do we have an option to define the maximum storage space for the database? E.g. 100MB, so that we don't run out of memory on Jetson

env/arm/Dockerfile
Rut-Vahab: The database already runs in a separate container (agstream_mongo).
Our ARM container only connects to it — it doesn’t host or run MongoDB itself.
So the setup already follows the separation principle between the app and the database.

docs/mongodb_setup.md
Rut-Vahab: we add to the docs how to push data and to check it .

About the memory limit - we don't have administrator permissions so we can't do that.

env/arm/Dockerfile
lyuzinmaxim: Why do we install it then in the container?

lyuzinmaxim (approved):

PR #77, opened 2025-10-23
Enable NVIDIA hardware decoding using nvv4l2decoder

Rut-Vahab:
Replaced CPU-based decodebin with NVIDIA hardware-accelerated nvv4l2decoder.  
Performed benchmark comparing both methods on RTSP stream performance.  
Added documentation with results and conclusions in docs/deepstream_hardware_decoding.md.


src/deepstream_pipeline.py
```diff
@@ -73,15 +74,10 @@ def on_pad_added(src, new_pad):
...
-    def on_decode_pad_added(src, new_pad):
-        sink_pad = convert.get_static_pad("sink")
-        if not sink_pad.is_linked():
-            new_pad.link(sink_pad)
+    
 
     rtspsrc.connect("pad-added", on_pad_added)
-    decode.connect("pad-added", on_decode_pad_added)
-
+    decode.link(convert)
```
Sarah5567: please move this line to the links section, after this line: `parse.link(decode)`

src/deepstream_pipeline.py
Sarah5567, line 1: please run linting (flake8, isort, black)

docs/deepstream_hardware_decoding.md
```diff
@@ -0,0 +1,67 @@
...
+decode = Gst.ElementFactory.make("nvv4l2decoder", "decode")
+```
+
+---
+
+## 3. Benchmark Results
+
+| Metric | CPU (`decodebin`) | GPU (`nvv4l2decoder`) |
+|--------|-------------------|------------------------|
+| Total runtime | 7.9 s | 159.4 s |
```
lyuzinmaxim: what do you mean by `total runtime`?

src/deepstream_pipeline.py
```diff
@@ -73,15 +75,9 @@ def on_pad_added(src, new_pad):
         sink_pad = depay.get_static_pad("sink")
         if not sink_pad.is_linked():
             new_pad.link(sink_pad)
-
```
lyuzinmaxim, line 93: what is the reason behind deleting those lines?

src/deepstream_pipeline.py
```diff
@@ -73,15 +75,9 @@ def on_pad_added(src, new_pad):
...
-    def on_decode_pad_added(src, new_pad):
-        sink_pad = convert.get_static_pad("sink")
-        if not sink_pad.is_linked():
-            new_pad.link(sink_pad)
+    
 
     rtspsrc.connect("pad-added", on_pad_added)
-    decode.connect("pad-added", on_decode_pad_added)
```
lyuzinmaxim, line 100: same here

docs/deepstream_hardware_decoding.md
Rut-Vahab: It’s the total pipeline runtime measured on the test stream using `time python3 src/deepstream_pipeline.py`.

src/deepstream_pipeline.py
Rut-Vahab: These lines were removed because the dynamic pad linking is handled in `on_pad_added`, so they are redundant.

src/deepstream_pipeline.py
Rut-Vahab: These are required for dynamic pad handling; the pipeline runs successfully with them, as verified.

docs/deepstream_hardware_decoding.md
lyuzinmaxim: so the processing time is 20 times longer for hardware encoding?

Rut-Vahab: Yes
Both runs processed 5693 frames, but the GPU decoding took ~159 s vs ~7.9 s on CPU in this test.

lyuzinmaxim: That's super strange. In general, it should be faster than CPU decoding. Could you double check?

Rut-Vahab: Yes, you asked me, I checked again because it was very strange for me too, but the results were the same.

PR #79, opened 2025-10-26
Background removal techniques

### LINKED ISSUE #75: Investigate available background removal technics
As discussed earlier ([here](https://bootcamps2025.slack.com/archives/C09BW2TN7S8/p1758107001845019) and [here](https://bootcamps2025.slack.com/archives/C09BW2TN7S8/p1758051726202049)), and based on the way the segmentation is being used in the [WeedZSL repository](https://github.com/SyArsRa/WeedZSL/blob/71d4a281379299cc2146f43adf41e1d1a7d8d9d3/evaluate_embedding_models.py#L252), we decided to use this approach as the preprocessing step before the nvinfer inference.

There're another approaches of doing that except for using Excess Green Index. Please try to do the research and propose alternative approaches.

As the outcome of the ticket, please prepare the documentation where you provide code examples, and compare different approaches (based on libs requirements, complexity, possibility to port to GPU, visual quality with examples), and measure the CPU latency (e.g., on one of the test images)

DoD:
- Detailed readme (with examples) created

Sarah5567:
Adds a detailed README comparing vegetation segmentation methods used as preprocessing before `nvinfer`.

Includes:
- Code examples for six methods  
- Accuracy, latency, and GPU support comparison  
- Visual results and analysis  

Result: **ExG + Otsu + Morphology** recommended as the final preprocessing method (best accuracy).

closes #75


docs/background removal techniques.md
```diff
@@ -0,0 +1,502 @@
...
+![Accuracy Per Class Distribution](images/accuracy_per_class.png)
+
+Some classes remain at 0% accuracy, indicating that the model does not detect them at all - most likely because these classes were **not sufficiently represented during training**.
+
+**Analysis:**
+- The affected classes are vegetation types the model was not trained well on.  
+- Their exclusion from the evaluation led to a noticeable accuracy increase.  
+- This suggests that model retraining or data balancing is needed for better coverage.
+
+### Post-Optimization Results
```
lyuzinmaxim: Please do not include this section.

The classes we fail on should be documented, and we don't want to exclude them to get better accuracy numbers (it's an antipattern in general!).

We performed hierarchical evaluation without excluding any classes. This ensures that we can assess performance across broader categories - for example, misclassifying a two-leaf maize stage as a four-leaf stage is much less critical than misclassifying it as a weed

Sarah5567: But we can’t say that our model is able to classify those classes - its accuracy on them is zero.

lyuzinmaxim: @Sarah5567 and we don't have to say that we're able to classify then. We just need to document this as it is

Sarah5567: I'm not sure I understand you - is there a specific problem with the section that starts with `Post-Optimization Results`, or is the overall content of `Class-Specific Performance Issue` incorrect?

lyuzinmaxim (approved):

PR #86, opened 2025-10-27
Create  e2e benchmark proposal

### LINKED ISSUE #85: Create an e2e speed benchmark
We want to measure the end-to-end (E2E) performance of our DeepStream pipeline using the existing `src/deepstream_pipeline.py`.  
The goal is to benchmark **latency** and **throughput (FPS)** using a controlled RTSP input.

The script should:  
- Start an RTSP server that streams a local video file.  
- Launch the DeepStream pipeline (`deepstream_pipeline.py`) that consumes this stream.  
- Measure the end-to-end latency (input → output), based on the frames
   - **ADDED** Measure the averages of latencies for frame groups of 0...10, 10...100, 100...1000, 1000...EOF
- Calculate the average pipeline FPS after processing completes.

**NOTE** Before doing the implementation please create a PR with a proposal first.

**Tasks**  
- a tiny PR with proposal (as readme) for measuring speed is implemented (high-level overview)
- Create a Python script that coordinates both the RTSP server and the pipeline.  
- Collect timing information (e.g., frame timestamps, processing duration).  
- Compute and print latency statistics and average FPS.  
- Optionally, log the results to a file for comparison across runs.  

**Definition of Done (DoD)**  
- [ ] Python script launches RTSP stream and DeepStream pipeline automatically.  
- [ ] End-to-end latency is measured and printed.  
- [ ] Average FPS is computed after run completion.  
- [ ] Steps to run and results are documented in `docs/benchmark_e2e.md`.

Sarah5567:
This PR adds a proposal document (`docs/benchmark_proposal.md`) describing the planned approach for **end-to-end performance benchmarking** of the DeepStream pipeline.

The document outlines:
- Measurement strategy for latency and throughput (FPS)
- Planned workflow using GStreamer tracers
- Metrics and analysis methods
- References to NVIDIA guidelines

This is a high-level proposal only - no code changes are included.

Linked to #85


docs/benchmark_proposal.md
```diff
@@ -0,0 +1,101 @@
...
+- Frame drop rate
+- Processing efficiency (actual FPS vs. input FPS)
+
+## Best Practices (NVIDIA Guidelines)
+
+Following [NVIDIA's DeepStream performance measurement guidelines](https://docs.nvidia.com/metropolis/deepstream/dev-guide/text/DS_Performance.html):
+
+1. Set GPU to maximum performance mode
+2. Use consistent test conditions (same video, resolution, pipeline config)
+3. Include warm-up period (discard initial frames)
```
lyuzinmaxim: Let's not do the warmup thing - it's a bit out of scope

docs/benchmark_proposal.md
lyuzinmaxim, line 1: Do we need to use a script (python file) which includes all elements (rtsp stream, streammux, ..., nvmsgconv) for running that e2e benchmark (smth like `run_e2e_deepstream_pipeline.py`?)? Because for the pipeline itself, we have the `deepstream_pipeline.py` file. As soon as we change the `deepstream_pipeline.py`, we'd need to update the `run_e2e_deepstream_pipeline.py` (or whatever name).

Sarah5567: No, the idea is to run an independent script that launches both rtsp_server.py and deepstream_pipeline.py with the required environment variables that make the pipeline print logs about FPS and latency.
It would look something like this:
```python

with open(LOG_FILE, "w") as f:
    ds_proc = subprocess.Popen(
        [
            "bash", "-c",
            f"GST_TRACERS='latency;stats' GST_DEBUG='GST_TRACER:7' python3 {PIPELINE_PATH}"
        ],
        stdout=f, stderr=subprocess.STDOUT
    )
```
Since the output will be redirected to a log file rather than the console, the script will later analyze those logs to extract the required information.
 You can see an example of how this analysis is performed[here](https://github.com/Dorozhko-Anton/ComputerVisionStreams/blob/main/TracingLatency/deepstream_pipeline/parse_latency_trace.py).

lyuzinmaxim: So we don't have to change anything in our <PIPELINE_PATH> file to run the benchmark?

Sarah5567: yes

lyuzinmaxim (approved):

PR #89, opened 2025-10-30
Benchmark implementation

### LINKED ISSUE #85: Create an e2e speed benchmark

Sarah5567:
This PR adds `src/benchmark.py`, a script that benchmarks the end-to-end (E2E) performance of the DeepStream pipeline.

- Starts RTSP server → runs DeepStream pipeline → analyzes GStreamer tracer logs.  
- Computes average and grouped latencies (0–10, 10–100, 100–1000, 1000–EOF).  
- Calculates overall FPS and saves results to CSV files under `/workspace/output/`.

> Note: In rare cases, the pipeline may hang mid-run due to GStreamer thread sync issues.

closes #85


src/benchmark.py
```diff
@@ -0,0 +1,149 @@
...
+
+    try:
+        runtime_sec = run_deepstream_pipeline()
+    except KeyboardInterrupt:
+        print("Interrupted by user.")
+        runtime_sec = 0
+    finally:
+        print("Stopping RTSP server...")
+        rtsp_proc.terminate()
+        rtsp_proc.wait()
```
Rut-Vahab: Suggestion: add a timeout to prevent potential hanging when the RTSP process doesn't terminate gracefully.
Example:
    rtsp_proc.terminate()
    try:
        rtsp_proc.wait(timeout=5)
    except subprocess.TimeoutExpired:
        rtsp_proc.kill()

src/benchmark.py
```diff
@@ -0,0 +1,149 @@
...
+
+    analyze_logs(runtime_sec)
```
Rut-Vahab, line 194: Suggestion: call `os.makedirs(os.path.dirname(LATENCY_REPORT), exist_ok=True)` 
before writing CSVs, to ensure the output directory exists even if deleted.

src/benchmark.py
```diff
@@ -0,0 +1,149 @@
...
+        run_pipeline(pipeline)
+        end_time = time.time()
+        return end_time - start_time
+    except KeyboardInterrupt:
+        print("Received KeyboardInterrupt — stopping pipeline manually...")
+        raise
+    finally:
+        print("Pipeline finished execution.")
+        
+def get_video_frame_count(video_path):
```
Rut-Vahab, line 75: Suggestion: wrap `cap.release()` in a `finally` block to ensure the capture is always released,
even if an exception occurs during reading.

docs/benchmark_e2e.md
lyuzinmaxim, line 1: Could you please run this benchmark on a long video that we have?

env/arm/Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 RUN pip3 install --no-cache-dir --upgrade pip wheel setuptools
 
-RUN pip3 install --no-cache-dir isort black flake8 
+RUN pip3 install --no-cache-dir isort black flake8 opencv-python-headless
```
lyuzinmaxim: Once this PR is merged, please merge main in your branch, so you don't need those changes.

src/benchmark.py
lyuzinmaxim, line 1: Please add typings and run linting

src/benchmark.py
```diff
@@ -0,0 +1,154 @@
...
+    if not cap.isOpened():
+        print(f"Failed to open video: {video_path}")
+        return 0
+    try:
+        total_frames = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
+        return total_frames
+    finally:
+        cap.release()
+
+def analyze_logs(runtime_sec):
```
lyuzinmaxim: I think it's possible to split this function into several:
- opening file and parsing latencies
- calculate latencies per group etc (here I think it makes sense to use BaseModel from pydantic to have a clear overview of the structures we use (e.g., for groups, or/and for csv outputs)
- saving results from df to csv

docs/benchmark_e2e.md
Sarah5567: Currently, we can’t run any code that uses `NumPy` or `OpenCV` due to the issues with the Dockerfile.

lyuzinmaxim (approved):

Issue #87, opened 2025-10-28
Implement a modular background removal technique


lyuzinmaxim:
Based on the ticket #75 we've chosen the "ExG + Otsu + Morphology" method for the background removal. Also, based on the ticket #57 we've prepared the the way for accessing frames.

The goal of this ticket would be to create some kinda modular construction which can be used in a callback function - function/class/... .

The idea is to have all operations on GPU (if possible). To achieve this, please take a look at the following resources:
- [numpy (CPU implementation)](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-imagedata-multistream/deepstream_imagedata-multistream.py#L121)
- ~[cupy (same thing, but on GPU)](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-imagedata-multistream-cupy/deepstream_imagedata-multistream_cupy.py#L106), [its readme](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-imagedata-multistream-cupy/README#L85)~

~**NOTE**: If there's a need for opencv+cuda package, please raise this.~

~Please note that Otsu threshold operation has no GPU support. If we will not find a workaround in this ticket, we'll create a ticket to implement it.~

DoD:
- a "module" for background removal is created
- approach is tested on test images
- latencies and tradeoffs documented

lyuzinmaxim:
Please create a new simple pipeline showing the usecase of this operation, e.g. `deepstream_pipeline_background_removal.py`

PR #90, opened 2025-10-30
Add modular CPU-only background removal module using ExG + Otsu + Mor…

Rut-Vahab:
# Background Removal Module (CPU-only)

## Overview
This module implements background removal using the **Excess Green (ExG) + Otsu threshold + Morphological operations** approach.  
It is modular, ready to be used in a callback function or integrated into pipelines such as DeepStream.

## Features
- `remove_background_exg_otsu_morph(frame)`: Modular function to remove background from a single BGR frame.
- `run_demo()`: Simple demo pipeline to process all images in the test folder.
  - Saves both the binary mask and the resulting image.
  - Prints processing time (latency) for each image.
- CPU-only implementation ensures compatibility and stability on Jetson devices.
- Modular design allows easy integration into callbacks or other pipelines.

## Notes
- GPU acceleration using CuPy or OpenCV CUDA was considered.
- **Otsu thresholding does not support GPU**, so the current solution is CPU-only.
- CPU-only implementation satisfies the **DoD** and allows reliable testing.
- Latencies and trade-offs are documented via console output for each image.

## Usage
```python
from background_removal import remove_background_exg_otsu_morph

mask, result = remove_background_exg_otsu_morph(frame)

### CODE REVIEW

File: src/deepstream_pipeline_background_removal.py [hunk 1]
```diff
@@ -0,0 +1,106 @@
...
+    Apply background removal using:
+      - Excess Green (ExG)
+      - Otsu threshold
+      - Morphological closing/opening
+    CPU-only implementation with NumPy/OpenCV
+    Returns: mask, result
+    """
+
+    # Separate channels
+    b = image_bgr[:, :, 0].astype(np.float32)
```
Sarah5567: Why do you use `numpy` and `opencv`? The idea is to use libraries that work on the GPU.
Consider searching for the best threshold instead of using `Otsu thresholding`, so the frames remain in device memory. or you could try implementing `Otsu thresholding` yourselves.

src/deepstream_pipeline_background_removal.py
```diff
@@ -0,0 +1,106 @@
...
+Includes timing for latency measurement.
+"""
+
+import cv2
+import numpy as np
+import os
+from datetime import datetime
+import time
+
+# 📁 תיקיות קלט ופלט
```
Sarah5567: Please translate all the Hebrew comments into English

src/deepstream_pipeline_background_removal.py
Sarah5567, line 1: I think you have to demonstrate how to use it within a pipeline, as you can see [here](https://github.com/KamaTechOrg/AgStream/issues/87#issuecomment-3466557751), not just the basic function.

lyuzinmaxim, line 1: please rename this file to have "cpu" in its name

env/jetson/Dockerfile
```diff
@@ -28,15 +28,29 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     gstreamer1.0-plugins-ugly \
     gstreamer1.0-libav \
     libcairo2-dev pkg-config \
+    libjpeg-dev libpng-dev libtiff-dev \
```
lyuzinmaxim: same here

https://github.com/KamaTechOrg/AgStream/pull/89/files#r2484809086

env/jetson/Dockerfile
```diff
@@ -28,15 +28,29 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 
-RUN pip3 install --no-cache-dir isort black flake8 
+# Optional code-style tools
+RUN pip3 install --no-cache-dir isort black flake8
 
+# DeepStream Python binding
 COPY pyds-1.1.10-py3-none-linux_aarch64.whl /tmp/
 RUN pip3 install /tmp/pyds-1.1.10-py3-none-linux_aarch64.whl
 
+# CuPy + Python packages for background removal
```
lyuzinmaxim: same here

https://github.com/KamaTechOrg/AgStream/pull/89/files#r2484809086

env/jetson/Dockerfile
```diff
@@ -28,15 +28,29 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     gstreamer1.0-plugins-ugly \
     gstreamer1.0-libav \
     libcairo2-dev pkg-config \
+    libjpeg-dev libpng-dev libtiff-dev \
  && rm -rf /var/lib/apt/lists/*
 
+# Python tools
```
lyuzinmaxim: the comments are not needed IMO

src/deepstream_pipeline_background_removal.py
lyuzinmaxim: I resolve it in favor of #92

src/deepstream_pipeline_background_removal.py
lyuzinmaxim: Agree. Also please see here how to access frame on jetson: https://github.com/KamaTechOrg/AgStream/blob/38b3eb9493c1164480d10b4e029d4eb69f200ed5/src/access_raw_frames_pipeline.py#L22-L34

lyuzinmaxim, line 1: as discussed today on daily, please add the test inside a deepstream pipeline (a simple one)

env/jetson/Dockerfile
lyuzinmaxim, line 1: please merge main into your branch

src/deepstream_pipeline_background_removal.py
lyuzinmaxim, line 1: Please add typings.

Also, in this file you show that your method (`remove_background_exg_otsu_morph()`) is working for images when they are loaded using OpenCV imread, that's fine. Also the background operation is modular - so we can potentially just import this function and use as is.

However, the `background_removal_callback()` is now doing nothing - why do we need it?

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python3
```
lyuzinmaxim: not needed

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,137 @@
...
+
+gi.require_version("Gst", "1.0")
+
+RTSP_URL = "rtsp://127.0.0.1:8554/test"
+OUTPUT_DIR = f"/workspace/output/frames_{datetime.now().strftime('%Y%m%d_%H%M%S')}/"
+MASKED_DIR = os.path.join(OUTPUT_DIR, "masked/")
+os.makedirs(MASKED_DIR, exist_ok=True)
+
+
+def apply_vegetation_mask(image_path, output_path):
```
lyuzinmaxim: why do you apply this function if you could apply the one you've created (`remove_background_exg_otsu_morph`)?

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def on_message(bus, msg):
+        if msg.type == Gst.MessageType.ERROR:
+            err, _ = msg.parse_error()
+            print(f"❌ ERROR: {err}")
+            loop.quit()
+        elif msg.type == Gst.MessageType.EOS:
+            print("✅ End of stream")
+            loop.quit()
+        elif msg.type == Gst.MessageType.ELEMENT:
```
lyuzinmaxim: that's not the way it should be implemented. It should be done using a probe (callback), and we should use the function you created (`remove_background_exg_otsu_morph`).

Please use [this file](https://github.com/KamaTechOrg/AgStream/blob/main/src/frames_skipping_pipeline.py) as a reference:
- we define there `frame_skip_probe` which contains the modular function `should_skip_frame`
- we run a simple pipeline defined in `build_pipeline`
- we attach the probe to the streammux element ([link](https://github.com/KamaTechOrg/AgStream/blob/3c19d6874d3dbc01c562cb1cf00983801e2b7e3f/src/frames_skipping_pipeline.py#L139))

You'd need to create the same thing, the only difference would be to have your custom prone (e.g., `background_removal_probe()`), and connect it to the element (e.g., to the same src pad of streammux).

As for the probe, you need to use the mechanism for retrieve the frame from the surface as used [here](https://github.com/KamaTechOrg/AgStream/blob/3c19d6874d3dbc01c562cb1cf00983801e2b7e3f/src/access_raw_frames_pipeline.py#L22C1-L34C33)

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,200 @@
+"""
+RTSP → Decode (GPU nvv4l2decoder) → NVMM RGBA → System RGBA → videoconvert (CPU) → BGR → Probe
+Apply background removal (ExG + Otsu + Morphology) on CPU in the probe and save frames.
+
+Rationale: In the DeepStream L4T image, `avdec_h264` may be unavailable. We keep decode on GPU
+and move frames to System Memory before the probe, so the algorithm runs on CPU only.
+"""
```
lyuzinmaxim: We don't need this, please remove

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(self, out_dir: str, save_every: int):
+        self.out_dir = out_dir
+        self.save_every = max(1, save_every)
+        self.counter = 0
+
+
+def background_removal_probe(pad, info, u_data: _Ctx):
+    buf = info.get_buffer()
+    if not buf:
+        return Gst.PadProbeReturn.OK
```
lyuzinmaxim: should we return `Gst.PadProbeReturn.OK`? In the case if buffer isn't available, smth is wrong with a pipeline. Having this will lead to the silent error

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,200 @@
...
+
+    frame = _map_bgr_from_buffer(pad, buf)
+    if frame is None:
+        return Gst.PadProbeReturn.OK
```
lyuzinmaxim: https://github.com/KamaTechOrg/AgStream/pull/90/files#r2495026098

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,200 @@
...
+import numpy as np
+
+
+RTSP_URL = os.environ.get("RTSP_URL", "rtsp://127.0.0.1:8554/test")
+OUTPUT_DIR = f"/workspace/output/frames_{datetime.now().strftime('%Y%m%d_%H%M%S')}/"
+MASKED_DIR = os.path.join(OUTPUT_DIR, "masked/")
+SAVE_EVERY = int(os.environ.get("SAVE_EVERY", "1"))  # save every N-th frame
+MORPH_KERNEL = int(os.environ.get("MORPH_KERNEL", "34"))
+
+os.makedirs(MASKED_DIR, exist_ok=True)
```
lyuzinmaxim: there's no need now to parse the exported env var, especially for OUTPUT_DIR, MASKED_DIR, SAVE_EVERY, MORPH_KERNEL

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,200 @@
...
+        arr = np.frombuffer(map_info.data, dtype=np.uint8)
+        expected = height * stride
+        if arr.size >= expected:
+            frame2d = arr[:expected].reshape((height, stride))
+            if stride == width * 3:
+                frame = frame2d.reshape((height, width, 3))
+            else:
+                frame = frame2d[:, : width * 3].reshape((height, width, 3)).copy()
+        else:
+            frame = arr.reshape((height, width, 3))
```
lyuzinmaxim: ~did you encounter such cases? Aren't videos we use all the same in terms of their raw representation?~

please see https://github.com/KamaTechOrg/AgStream/pull/90/files#r2495080262

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,200 @@
...
+        return frame
+    finally:
+        buf.unmap(map_info)
+
+
+class _Ctx:
```
lyuzinmaxim: It's a not very nice class name. Please consider having the name which is clear when you see it the first time

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(self, out_dir: str, save_every: int):
+        self.out_dir = out_dir
+        self.save_every = max(1, save_every)
+        self.counter = 0
+
+
+def background_removal_probe(pad, info, u_data: _Ctx):
```
lyuzinmaxim: please add typings

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -0,0 +1,200 @@
...
+    exg_norm = cv2.normalize(exg, None, 0, 255, cv2.NORM_MINMAX).astype(np.uint8)
+    _thr, mask = cv2.threshold(exg_norm, 0, 255, cv2.THRESH_BINARY + cv2.THRESH_OTSU)
+    kernel = cv2.getStructuringElement(cv2.MORPH_ELLIPSE, (MORPH_KERNEL, MORPH_KERNEL))
+    mask = cv2.morphologyEx(mask, cv2.MORPH_CLOSE, kernel)
+    out = np.zeros_like(frame)
+    out[mask > 0] = frame[mask > 0]
+    return out
+
+
+def _map_bgr_from_buffer(pad: Gst.Pad, buf: Gst.Buffer) -> np.ndarray | None:
```
lyuzinmaxim: I think this function is too complex. Please see how it's implemented [here](https://github.com/KamaTechOrg/AgStream/blob/bd51104625ac43211bba1dc50d5811c53e663dfe/src/access_raw_frames_pipeline.py#L22) - in the ticket description I also attached the same reference ([link](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-imagedata-multistream/deepstream_imagedata-multistream.py#L121))

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -51,43 +52,43 @@ def _map_bgr_from_buffer(pad: Gst.Pad, buf: Gst.Buffer) -> np.ndarray | None:
...
+        self.counter: int = 0
 
 
-def background_removal_probe(pad, info, u_data: _Ctx):
+def background_removal_probe(pad: Gst.Pad, info: Gst.PadProbeInfo, u_data: BackgroundRemovalCtx) -> Gst.PadProbeReturn:
     buf = info.get_buffer()
     if not buf:
-        return Gst.PadProbeReturn.OK
+        print("[probe][error] missing GstBuffer; dropping")
+        return Gst.PadProbeReturn.DROP  # avoid silent OK
```
lyuzinmaxim: I checked the docs, and there's no way to signal the error: https://gstreamer.freedesktop.org/documentation/gstreamer/gstpad.html?gi-language=c#GstPadProbeReturn

can we then raise an exception/error?

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -51,43 +52,43 @@ def _map_bgr_from_buffer(pad: Gst.Pad, buf: Gst.Buffer) -> np.ndarray | None:
...
 
     frame = _map_bgr_from_buffer(pad, buf)
     if frame is None:
-        return Gst.PadProbeReturn.OK
+        print("[probe][error] failed to map frame; dropping")
```
lyuzinmaxim: It's better to throw an error/exception

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -41,8 +42,8 @@ def remove_background_exg_otsu_morph(frame: np.ndarray) -> np.ndarray:
     return out
 
 
-def _map_bgr_from_buffer(pad: Gst.Pad, buf: Gst.Buffer) -> np.ndarray | None:
-    """Return BGR HxWx3 numpy array from a CPU buffer. Handles stride via GstVideoMeta.
+def _map_bgr_from_buffer(pad: Gst.Pad, buf: Gst.Buffer) -> Optional[np.ndarray]:
+    """Return BGR HxWx3 numpy array from a CPU buffer using GstVideoMeta stride.
```
lyuzinmaxim: All these operations are not needed - we should use here pyds function for that. I already mentioned it here: https://github.com/KamaTechOrg/AgStream/pull/90#discussion_r2495080262. Please address it

lyuzinmaxim (approved):
Approved. Please merge main first into your branch

PR #93, opened 2025-11-02
Frames skipping implementation

### LINKED ISSUE #88: Implement frame skipping technique based on frames comparison
We previously evaluated several CPU-based frame comparison methods and confirmed that frames can be skipped using GStreamer pad probes (#70).  
Now, we want to implement the chosen algorithm (#81 ) in a **modular way ** so it can be integrated into the main DeepStream pipeline later.

The idea is to have all operations on GPU (if possible) to avoid GPU->CPU->GPU copies. To achieve this, please take a look at the following resources:
- [numpy (CPU implementation)](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-imagedata-multistream/deepstream_imagedata-multistream.py#L121)
- ~[cupy (same thing, but on GPU)](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-imagedata-multistream-cupy/deepstream_imagedata-multistream_cupy.py#L106), [its readme](https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-imagedata-multistream-cupy/README#L85)~

**NOTE**: If there's a need for opencv+cuda package, please raise this.

**Tasks**  
- Investigate the possibilities to run the frame comparison algorithm on GPU (cupy is preferred)
- Implement the frame comparison logic based on the selected algorithm from #81 .  
- Integrate the logic into a modular component or plugin (to be reusable).  
- Provide a simple demo/test pipeline to verify frame comparison is running successfully (see the `docs/frame_comparison_results.md`)

DoD:
- A modular approach for skipping frames based on the outcome of the ticket #81 is implemented

Sarah5567:
**Summary**
Implemented a modular frame comparison component (`FrameChangeDetector`) for selective frame processing in the DeepStream pipeline.  
The logic was integrated into a deepstream pipeline using a pad probe that drops similar frames based on MSE, SSIM, and Optical Flow thresholds.

**Changes**
- Added modular CPU-based frame comparison logic in `frame_change_detector.py`.
- Integrated `FrameChangeDetector` into `deepstream_pipeline.py` via `frame_skip_probe`.
- Verified on sample RTSP video - the pipeline now processes **~21.46%** of frames (skipping similar ones).

> **Note:** currently implemented **on CPU only** -GPU implementation (`cv2.cuda` / `cupy`) pending due to installation issues.

closes #88


src/frames_skipping_pipeline.py
```diff
@@ -0,0 +1,195 @@
...
+    gst_buffer = info.get_buffer()
+    if not gst_buffer:
+        print("Unable to get GstBuffer")
+        return Gst.PadProbeReturn.OK
+
+    batch_meta = pyds.gst_buffer_get_nvds_batch_meta(hash(gst_buffer))
+    if not batch_meta:
+        return Gst.PadProbeReturn.OK
+
+    pyds.nvds_acquire_meta_lock(batch_meta)
```
r83575: Using `nvds_acquire_meta_lock `despite read-only metadata access.

src/frames_skipping_pipeline.py
```diff
@@ -0,0 +1,195 @@
...
+    l_frame = batch_meta.frame_meta_list
+    if l_frame is not None:
+        try:
+            frame_meta = pyds.NvDsFrameMeta.cast(l_frame.data)
+            return should_skip_frame(info, frame_meta, batch_meta)
+        except StopIteration:
+            pass
+        finally:
+            pyds.nvds_release_meta_lock(batch_meta)
```
r83575: Using `nvds_release_meta_lock` despite read-only metadata access.

env/jetson/Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 RUN pip3 install --no-cache-dir --upgrade pip wheel setuptools
 
-RUN pip3 install --no-cache-dir isort black flake8 
+RUN pip3 install --no-cache-dir isort black flake8 opencv-python-headless numpy
```
lyuzinmaxim: Please merge main - I suppose this is not needed

src/frames_skipping_pipeline.py
```diff
@@ -0,0 +1,191 @@
...
+# configurations
+RTSP_URL = "rtsp://127.0.0.1:8554/test"
+CONFIG_FILE = "/workspace/src/configs/resnet18.txt"
+OUTPUT_DIR = "/workspace/output/frames/"
+MESSAGE_TEMPLATE = "Frame processed: {}"
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+detector = FrameChangeDetector()
+
+def should_skip_frame(info, frame_meta, batch_meta):
```
lyuzinmaxim: Frame change detector you've initialized (`detector = FrameChangeDetector()`) is not in the context of the function - so once we import the function from another file, it will be gone.

Please check how to make it accessible

src/frames_skipping_pipeline.py
lyuzinmaxim, line 1: please run linting and add typings

src/frames_skipping_pipeline.py
Sarah5567: Do you think it's a proper approach to implement a singleton in `FrameChangeDetector`?

lyuzinmaxim: `frame_skip_probe` runs on each frame kinda indepentently. If you want to access the previous frame, you'd need to have an object which is still "alive" (and which contains the prev frame in our case). Singleton is one of the approaches, yes

env/jetson/Dockerfile
```diff
@@ -115,7 +115,7 @@ RUN mkdir -p /tmp/opencv/build \
...
```
lyuzinmaxim: It's not needed. Please see this thread: https://bootcamps2025.slack.com/archives/C09BW2TN7S8/p1762165919662759

src/frame_change_detector.py
```diff
@@ -33,24 +63,25 @@ def optical_flow_diff(self, imgA: NDArray[np.uint8], imgB: NDArray[np.uint8]) ->
         grayB = cv2.cvtColor(imgB, cv2.COLOR_BGR2GRAY)
         flow = cv2.calcOpticalFlowFarneback(grayA, grayB, None, 0.5, 3, 15, 3, 5, 1.2, 0)
         mag, _ = cv2.cartToPolar(flow[..., 0], flow[..., 1])
-        return float(np.mean(mag))
+        active = mag[mag > 0.5]
```
lyuzinmaxim: I see you added new logic. Please add the "0.5" constant to other contants in this file. Also please add the steps you perform to the existing frame comparison readme

src/frame_change_detector.py
```diff
@@ -2,14 +2,44 @@
...
+FLOW_THRESH = 0.7
+MIN_STATIC_FRAMES = 4
+
 
 class FrameChangeDetector:
-    def __init__(self, mse_thresh: float = 50.0, ssim_thresh: float = 0.995, flow_thresh: float = 0.1) -> None:
+    _instance = None  # singleton instance
+
+    def __new__(cls, *args, **kwargs):
+        # Create only one instance per process
```
lyuzinmaxim, line 17: even though here you're checking whether this object was initialized, are you sure that this is actually skipping the frames, i.e. working as expected?

Because here we're initializing new object every frame:
```
detector = FrameChangeDetector()
```

src/frames_skipping_pipeline.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
+"""
+DeepStream Pipeline
+RTSP → Decode → Convert → Save Images
+"""
```
lyuzinmaxim: it's not needed

src/frames_skipping_pipeline.py
lyuzinmaxim, line 1: please add typings

src/frame_change_detector.py
Sarah5567: Yes, it really skips frames

src/frames_skipping_pipeline.py
```diff
@@ -0,0 +1,177 @@
...
+from frame_change_detector import FrameChangeDetector
+
+# configurations
+RTSP_URL = "rtsp://127.0.0.1:8554/test"
+CONFIG_FILE = "/workspace/src/configs/resnet18.txt"
+OUTPUT_DIR = "/workspace/output/frames/"
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+
+def should_skip_frame(info: Any, frame_meta: Any, batch_meta: Any) -> int:
```
lyuzinmaxim, line 21: just as a double-check, could you please create a file called `frame_skipping_callback.py`, and move `should_skip_frame` and `frame_skip_probe` to there, and import those callbacks from it in this file?

src/frames_skipping_pipeline.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env python3
```
lyuzinmaxim: it's not needed

env/jetson/Dockerfile
```diff
@@ -115,7 +115,7 @@ RUN mkdir -p /tmp/opencv/build \
 
 RUN pip3 install --no-cache-dir --upgrade pip wheel setuptools
 
-RUN pip3 install --no-cache-dir isort black flake8 
+RUN pip3 install --no-cache-dir isort black flake8
```
lyuzinmaxim: no need to change

src/frames_skipping_pipeline.py
Sarah5567: I did that, still skips frames.

Sarah5567: This code implements the Singleton pattern because when `FrameChangeDetector()` is called, `__new__` is executed first and checks if `_instance` is `None`. If it is, a new object is created and stored in the class-level variable `_instance`. (I saw a similar implementation [here](https://www.geeksforgeeks.org/python/singleton-pattern-in-python-a-complete-guide/#:~:text=reflected%20in%20samplemodule2.-,Classic%20Singleton%3A,-Classic%20Singleton%20creates)).
After that, `__init__` uses `_initialized` to prevent reinitialization of this single instance.

lyuzinmaxim (approved):

PR #94, opened 2025-11-02
feat: Add OpenCV+CUDA support

lyuzinmaxim:
According to the nvidia [documentation](https://docs.nvidia.com/metropolis/deepstream/6.4/dev-guide/text/DS_docker_containers.html#a-docker-container-for-jetson), there're two images available for deepstream 6.4:
- `deepstream:6.4-samples-multiarch` - contains **the runtime libraries** (suitable for running production code)
- `deepstream:6.4-triton-multiarch` - contains **devel libraries** (contains the cuda-related binaries, e.g. nvcc compiler)

It was not possible to compile OpenCV with CUDA support while using the `deepstream:6.4-samples-multiarch` image. Changing the base image resolved this


env/jetson/Dockerfile
```diff
@@ -30,6 +33,86 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+    && cd /tmp/cmake \
+    && wget https://github.com/Kitware/CMake/releases/download/v3.27.7/cmake-3.27.7.tar.gz \
+    && tar -xf cmake-3.27.7.tar.gz \
+    && cd cmake-3.27.7 \
+    && ./bootstrap \
+    && make -j "$(nproc)"\
+    && make install \
+    && rm -R /tmp/cmake
+
+RUN OPENCV_VERSION=4.11.0 && cd /tmp && \
```
r83575, line 46: Move `OPENCV_VERSION=4.11.0` to an `ARG` at the top of the Dockerfile.  
This improves clarity and makes the version easily configurable without touching build layers.

env/jetson/Dockerfile
```diff
@@ -30,6 +33,86 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+    -D WITH_LAPACK=NO \
+    -D WITH_OPENGL=OFF \
+    -D WITH_OPENMP=ON \
+    -D WITH_QT=OFF \
+    -D WITH_TBB=OFF \
+    -D WITH_V4L=OFF .. \
+    -D OPENCV_GENERATE_PKGCONFIG=YES \
+    && cd /tmp/opencv/build \
+    && make -j $(nproc --all) \
+    && make -j $(nproc --all) install \
```
r83575, line 111: There are two consecutive `make` calls,
Combine them or run make once followed by make install. The first build call is redundant.

env/jetson/Dockerfile
```diff
@@ -30,6 +33,86 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+    -D BUILD_opencv_java=OFF \
+    -D BUILD_opencv_python2=OFF \
+    -D BUILD_opencv_python3=ON \
+    -D BUILD_PERF_TESTS=OFF \
+    -D BUILD_TESTS=OFF \
+    -D BUILD_TIFF=ON \
+    -D INSTALL_C_EXAMPLES=OFF \
+    -D INSTALL_PYTHON_EXAMPLES=OFF \
+    -D OPENCV_EXTRA_MODULES_PATH=/tmp/opencv_contrib/modules \
+    -D PYTHON_DEFAULT_EXECUTABLE=$(which python3.10) \
```
r83575, line 90: Using `python3.10` directly may cause breakage if the base image changes.
Make this dynamic using `python3 -c` or expose as a configurable `ARG`.

r83575 (commented):
These are minor improvements and don’t block the PR, but implementing them would increase `Dockerfile` maintainability, reduce size, and improve robustness.

env/jetson/Dockerfile
lyuzinmaxim: I put it in here for simplicity. I see your point - same for cmake installation.

env/jetson/Dockerfile
lyuzinmaxim: It's a part of opencv compilation, please see the docs, section "Install":
https://docs.opencv.org/4.11.0/d7/d9f/tutorial_linux_install.html

env/jetson/Dockerfile
lyuzinmaxim: We are still super dependant on the base image provided by nvidia - so anyway we'd have to change the python version in here. I'd just keep this explicitly to that we point to the exact python version we use

PR #101, opened 2025-11-05
Integrate unique RTSP servers for each user

### LINKED ISSUE #100: Create and integrate unique RTSP servers for each of the users
Currently, everyone who is using the repository is using or reusing the same RTSP stream. It's not a safe behaviour. In this ticket, we need to set up unique port for RTSP stream for each of the users, based on the user id.

In run_dev_jetson.sh script, we need to expose unique ports, and need to export the env var (e.g., `RTSP_PORT`).

In rtsp_server, use this env var (`RTSP_PORT`) to set up the stream.
In all RTSP consumer scripts, please change the hardcoded port (8554) to use the env var (`RTSP_PORT`)

DoD:
- RTSP streams are unique for each of the users
- all scripts dependent on RTSP are updated to work in a new setup.

Rut-Vahab:
### RTSP Port Isolation per User

This update ensures each user gets a unique RTSP port based on their user ID.  
Changes include:
- Updated `run_dev_jetson.sh` to expose and export a unique `RTSP_PORT`.  
- Modified `rtsp_server.py` to read the port from the environment variable.  
- Updated `deepstream_pipeline.py` to use `RTSP_PORT` instead of a hardcoded value.  

✅ Each user's RTSP stream now runs independently and safely.

> **Note:**  
> After merging this change into `main`, all pipelines must update their RTSP access method.  
> Replace:  
> ```python
> RTSP_URL = "rtsp://127.0.0.1:8554/test"
> ```  
> with:  
> ```python
> rtsp_port = os.environ.get("RTSP_PORT", "8554")
> RTSP_URL = f"rtsp://127.0.0.1:{rtsp_port}/test"
> ```
> to ensure compatibility with the new per-user RTSP port setup.

closes #100


run_dev_jetson.sh
```diff
@@ -26,13 +31,14 @@ docker build \
...
 
-# Start container (detached mode, no terminal)
+# Start container (detached mode)
 docker run \
     --name ${CONTAINER_NAME} \
     -v ${workspace_dir}:/workspace \
     -u ${USER_ID}:${USER_ID} \
     --shm-size 8G \
-    -p ${USER_ID}:8554 \
+    -p ${RTSP_PORT}:8554 \
```
Sarah5567: Why did you change it?

src/deepstream_pipeline.py
```diff
@@ -12,7 +12,11 @@
 
 from gi.repository import GLib, Gst
 
-RTSP_URL = "rtsp://127.0.0.1:8554/test"
+# RTSP_URL = "rtsp://127.0.0.1:8554/test"
```
Sarah5567: Please remove this line.

run_dev_jetson.sh
r83575: - `RTSP_PORT` is calculated as `8554 + USER_ID`, making it a valid and unique TCP port per user.
- This ensures each user gets a dedicated RTSP stream without port conflicts on localhost, as required in the task.

Sarah5567: On second thought, you do need to change this line, but in a different way.
Since this line maps `host_port:container_port`, it should be changed to `-p ${USER_ID}:${RTSP_PORT}`.

r83575: The `-p` flag in Docker maps `host_port:container_port`.

Inside the container, the RTSP service always listens on port 8554.

`USER_ID` is not a valid or appropriate port number (e.g., it's not in the allowed port range).  
Instead, `RTSP_PORT` is calculated as `8554 + USER_ID`, giving each user a unique port on the host.

So `-p ${RTSP_PORT}:8554` is the correct and safe mapping to expose a dedicated RTSP stream per user, while keeping the internal container configuration unchanged.

Sarah5567: `USER_ID` worked well for us as a port on the host, but more importantly - the RTSP service no longer listens on port 8554. This is exactly the change that this ticket made, isn’t it?

r83575: You're right – after checking the server code, the RTSP service now listens on `RTSP_PORT` inside the container as well, not on 8554 anymore.

So the correct Docker port mapping should be:
`-p ${RTSP_PORT}:${RTSP_PORT}
`
Thanks for pointing that out – updating accordingly.

lyuzinmaxim (approved):

PR #102, opened 2025-11-05
Create helper function for checking/creating *.engine files

### LINKED ISSUE #99: Create a helper function for checking/creating *.engine files
Currently, in the model config file for nvinfer we provide the `model-engine-file` location ([here](https://github.com/KamaTechOrg/AgStream/blob/bd51104625ac43211bba1dc50d5811c53e663dfe/src/configs/mobilenet.txt#L10)). However, the logic of creating those files is in [this script](https://github.com/KamaTechOrg/AgStream/blob/main/src/test_onnx_pipeline.py) only. The idea of this ticket is to have a helper function which will ensure the *.engine files exist (and otherwise do the conversion) in the location specified in <model>.txt file, and to use this helper in every deepstream script as a "pre-launch" thing.

DoD:
- the helper function is integrated in every deepstream script, and is working seamlessly

Rut-Vahab:
# Add helper for generating TensorRT engine files

This PR adds `ensure_engine_exists()` which checks if the `.engine` file from the model config exists, and creates it from `.onnx` if missing.  
The function is called at the start of the DeepStream pipeline to guarantee the engine is ready before inference.

**Note:**  
After merging this change into `main`, all DeepStream pipelines must call `ensure_engine_exists(CONFIG_FILE)` at the beginning of their script, before building the pipeline.  

Closes #99


src/engine_helper.py
lyuzinmaxim, line 1: please add typings

src/engine_helper.py
```diff
@@ -0,0 +1,33 @@
...
+    onnx_path = None
+
+    with open(config_path, "r") as f:
+        for line in f:
+            if line.strip().startswith("model-engine-file"):
+                engine_path = line.split("=")[1].strip()
+            elif line.strip().startswith("onnx-file"):
+                onnx_path = line.split("=")[1].strip()
+
+    if not engine_path or not onnx_path:
```
lyuzinmaxim: please be more specific and raise different errors for onnx/deepstream

src/engine_helper.py
```diff
@@ -0,0 +1,33 @@
...
+
+    cmd = [
+        "trtexec",
+        f"--onnx={onnx_path}",
+        f"--saveEngine={engine_path}",
+        "--fp32",
+        "--workspace=2048"
+    ]
+    subprocess.run(cmd, check=True)
+    print(f"[INFO] Engine file created successfully: {engine_path}")
```
lyuzinmaxim, line 58: Did you test that engines you create like this are working?


You can just call this helper, and compare the *.engine files generated by it, and *.engine files generated by deepstream w/o this helper. You can use some hash functions to compare the content of the files, e.g. sha256

Rut-Vahab: ✅ Verified the engine generation and pipeline execution:
- Deleted previous .engine and regenerated successfully.
- SHA256 matches across runs.
- DeepStream loads the engine and runs inference.
- Frames are saved to the output directory as expected.
- Performance metrics within expected range.

lyuzinmaxim: @Rut-Vahab Regarding the model consistency I meant the following: we need to make sure the models we generate **without** this helper (basically, implicitly by deepstream when we run the pipeline first time) are identical to the models we get using `trtexec`. Could you please compare those ones?

Rut-Vahab: I compared the engine files generated by the helper (using trtexec) with the ones DeepStream generates implicitly on first run. I verified their SHA256 hashes and they match exactly, so the engines are identical.

PR #111, opened 2025-11-09
Gpu frame comparison 96

### LINKED ISSUE #96: Move frame comparison technique ops to GPU
In #88, we created the pipeline which demostrates the modular approach to skip the frames based on the comparison of consecutive ones. In this ticket, we just need to move frame comparison operations to GPU, and make sure that we don't do the GPU->CPU->GPU copies.

Please use OpenCV+CUDA instead of cupy.

Links:

- How to access raw GPU frames (directly, w/o copying to CPU): https://docs.nvidia.com/metropolis/deepstream/6.4/dev-guide/text/DS_sample_custom_gstream.html#accessing-nvbufsurface-memory-in-opencv
- The only way to access the GPU surface (directly) on Jetson devices: https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/bindings/docstrings/functionsdoc.h#L571

DoD:
- a new script is created (based on CPU script)
- frame comparison is working in the same way on GPU as on CPU

chani0343:
## GPU-Accelerated Frame Change Detection for DeepStream

Implements intelligent frame skipping using GPU-only processing to optimize DeepStream pipeline performance.

###  Key Features
- **GPU-Only Processing**: Direct CUDA-EGL integration with DeepStream NvBufSurface
- **Multi-Algorithm Detection**: MSE, SSIM, and Optical Flow on GPU
- **Optimized Thresholds**: Achieves 86% frame skip ratio (7/50 frames processed)
- **Memory Safe**: Proper GPU memory management and cleanup
- **C++ Performance**: pybind11 integration for maximum efficiency

Closes #96


src/gpu_frames_skipping_pipeline.py
lyuzinmaxim, line 1: please add typings

src/gpu_frames_skipping_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+
+# Configurations
+RTSP_URL = "rtsp://127.0.0.1:8554/test"
+CONFIG_FILE = "/workspace/src/configs/resnet18.txt"
+OUTPUT_DIR = "/workspace/output/frames/"
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+stats = {"total": 0, "skipped": 0, "processed": 0}
+
+def get_jetson_frame(gst_buffer, batch_id):
```
lyuzinmaxim: are you using this function? I am a bit confused since to me it seems like it's not called anywhere

src/gpu_frames_skipping_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+        if stats["total"] > 0:
+            skip_ratio = stats["skipped"] / stats["total"]
+            print(f"\n🔥 Jetson Final Stats:")
+            print(f"   Total: {stats['total']}")
+            print(f"   Processed: {stats['processed']}")
+            print(f"   Skipped: {stats['skipped']}")
+            print(f"   Skip Ratio: {skip_ratio:.1%}")
+
+if __name__ == "__main__":
+    run_jetson_pipeline()
```
lyuzinmaxim: please add a newline

PR #147, opened 2025-11-12
Drop manual frame saving in probe, use multifilesink as output in CPU

r83575:
This PR removes manual frame saving from the `background_removal_probe` and switches the pipeline's final sink to `multifilesink`.

- Dropped in-probe `cv2.imwrite` logic.
- Added `multifilesink` as the final pipeline element to persist output frames.


lyuzinmaxim (commented):
Did you check that there are frames in the output folder?
Btw, I don't see where you specify it

src/deepstream_pipeline_background_removal_cpu.py
```diff
@@ -70,12 +67,6 @@ def background_removal_probe(pad: Gst.Pad, info: Gst.PadProbeInfo, u_data: Backg
 
         masked = remove_background_exg_otsu_morph(frame_bgr)
```
Sarah5567: You need to write this modified frame into the buffer - it should look something like:
`np.copyto(np.array(surface, copy=False, order="C"), masked)`

lyuzinmaxim: I am not sure it's needed. Please confirm whether it's working now (without this command)

r83575: I tested it - this line was indeed needed. Without it, the changes to the frame were not propagated back to the Gst buffer. Adding the `np.copyto(...)` writes the modified frame back into the pipeline.

lyuzinmaxim (approved):

PR #149, opened 2025-11-13
Move bg removal to gpu

### LINKED ISSUE #92: Move background removal operations to work on GPU
Based on #87 , we need to use OpenCV+CUDA build in order to run all operations on GPU. Currently, we have modular constructions (func/class) that work on CPU. We need to have a file next to the CPU implementation with a "gpu" in its name. The logic should be exactly the same except for the operations ported to GPU.

As a reference for OpenCV+CUDA usage for raw frame access, please use the PR #96  .

For simplicity, I think we can port all background-removal logic to a probe defined in C++, and in python just link this probe using pybind.

Please note that some algorithms (Otsu) might be not working (please replace them by the other threshold operations.

DoD:
- A modular function for background removal `remove_background_gpu_probe(...)` is working

r83575:
Implements GPU-accelerated background removal pipeline using OpenCV+CUDA operations as requested in the task requirements.

## Changes Made
- **GPU Pipeline**: `deepstream_pipeline_background_removal_gpu.py` - GStreamer pipeline with GPU probe integration
- **C++ GPU Probe**: `gpu_background_probe.cpp` - OpenCV+CUDA implementation of ExG background removal
- **Compiled Module**: `gpu_background_probe.cpython-310-aarch64-linux-gnu.so` - pybind11 wrapper for Python integration
- **Additional GPU utilities**: `gpu_background_removal.cpp/.hpp` - Supporting GPU processing functions
- **Build System**: `Makefile` - Build configuration for GPU modules

## Technical Implementation
- **OpenCV+CUDA Operations**: ExG algorithm ported to GPU using `cv::cuda::` functions
- **Threshold Replacement**: Replaced Otsu with fixed threshold (127) for GPU compatibility
- **In-place Processing**: Direct buffer modification using CUDA graphics resources
- **pybind11 Integration**: C++ probe callable from Python pipeline
- **Same Logic as CPU**: Identical background removal algorithm but GPU-accelerated

## Testing
- GPU pipeline processes frames in-place on GPU memory
- Multifilesink saves processed frames at pipeline end
- Compatible with existing DeepStream infrastructure

Closes #92


lyuzinmaxim (commented):
The file `src/deepstream_pipeline_background_removal_cpu.py` doesn't need to be changed, right? Because we have it already in main.


Also I'd ask to restructure the code a bit to have a "module" for this.

```
src/
    deepstream/
        probes/
            background_removal/
                gpu/
                    gpu_background_probe.cpp
                    gpu_background_probe.cpython
                    gpu_background_removal.cpp
                    gpu_background_removal.hpp
        pipelines/
            deepstream_pipeline_background_removal_gpu.py
```

r83575:
You're right – the CPU file doesn't actually need to be changed. The reason it appears here is that originally I was asked to fix the CPU logic, and that work happened on the same branch before continuing with the GPU changes. So when I pushed the GPU-related commits, the existing CPU file changes were included again as part of the branch history.

lyuzinmaxim (approved):

PR #155, opened 2025-11-15
Fix pipelines

lyuzinmaxim:


configs/resnet18.txt
```diff
@@ -12,18 +12,13 @@ labelfile-path=/workspace/configs/crop_and_weed_83_classes.txt
...
 
-# 0: Detector, 1: Classifier, 2: Segmentation
-network-type=1
-
-classifier-threshold=0.5
-
-# 1: Primary, 2: Secondary
-process-mode=1
+# 0: Detector, 1: Classifier, 2: Segmentation, 100: Custom
+network-type=100
```
Sarah5567, line 21: Why did you change it?

src/deepstream/helpers/meta_tensor_extractor.py
Sarah5567, line 1: Is this approach for getting the predictions better than [this one](https://github.com/KamaTechOrg/AgStream/blob/497f96517c1e6900d746cf170a921d024ce60262/src/deepstream/pipelines/access_metadata.py#L56C1-L83C1)?

src/nvmsgbroker_pipeline.py
```diff
@@ -1,190 +1,260 @@
...
+    prediction: ClassificationPrediction
+
+
+class NvdsPlantEventBuilder:
+    """
+    For the reference, please check
+    https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/blob/9b27f02ffea46a3ded2ad26b3eea27ef3e2dfded/apps/deepstream-test4/deepstream_test_4.py
+    """
+
+    def build(self, batch_meta, frame_meta, event: PlantEvent) -> None:
```
Sarah5567, line 48: I tried something similar:
```Python

def create_and_add_message_metadata(batch_meta, frame_meta, frame_number: int, classification_results):
    """Create nvmsgbroker metadata and add it to frame metadata."""
    user_event_meta = pyds.nvds_acquire_user_meta_from_pool(batch_meta)
    if not user_event_meta:
        return

    msg_meta = pyds.alloc_nvds_event_msg_meta(user_event_meta)

    msg_meta.frameId = frame_number
    msg_meta.sensorId = 0
    msg_meta.placeId = 0
    msg_meta.moduleId = 0

    msg_meta.ts = pyds.alloc_buffer(32)
    pyds.generate_ts_rfc3339(msg_meta.ts, 32)

    person_obj = pyds.alloc_nvds_person_object()

    person_obj.age = 0

    msg_meta.extMsg = person_obj
    msg_meta.extMsgSize = sys.getsizeof(pyds.NvDsPersonObject)

    # Final attachment to frame
    user_event_meta.user_meta_data = msg_meta
    user_event_meta.base_meta.meta_type = pyds.NvDsMetaType.NVDS_EVENT_MSG_META
    pyds.nvds_add_user_meta_to_frame(frame_meta, user_event_meta)
```

Do you have any idea why your code works but mine doesn’t? Maybe my issue was deeper.

configs/resnet18.txt
lyuzinmaxim: I didn't check the source code of nvinfer but I suppose that specifying the `network-type` could've affect the way deepstream handles implicitly the metadata.

From what I understand, for classifiers deepstream expects already normalized tensor of probabilities (i.e. after softmax), which is not the case for us (our model just outputs raw logits). 

I was already using this `network-type=100` for other models where I was handling all postprocessing on my own (manually) using `output-tensor-meta` thing

src/deepstream/helpers/meta_tensor_extractor.py
lyuzinmaxim: I don't know whether it's better - my idea was to use directly the `NVDSINFER_TENSOR_OUTPUT_META`.

There's the following hierarchy:
- batch level
  - frame level
    - object level

We are using our models on the whole frame, so we are on frame level. However, to store the classification inference results on frame level, deepstream creates 1 "artificial" object (even though we don't have a detector model) per frame. The link you've sent shows how to parse this "artificial" object (created by deepstream as a workaround):
```
l_obj = frame_meta.obj_meta_list
obj_meta = pyds.NvDsObjectMeta.cast(l_obj.data)
l_classifier = obj_meta.classifier_meta_list
classifier_meta = pyds.NvDsClassifierMeta.cast(l_classifier.data)
```

Since we're using `output-tensor-meta` option for our models, we can retrieve model's outputs directly from the associated `NVDSINFER_TENSOR_OUTPUT_META` metadata (as I did): 
```
l_user = frame_meta.frame_user_meta_list
user_meta = pyds.NvDsUserMeta.cast(l_user.data)
if user_meta.base_meta.meta_type == pyds.NvDsMetaType.NVDSINFER_TENSOR_OUTPUT_META:
    return pyds.NvDsInferTensorMeta.cast(user_meta.user_meta_data)
```

In this case we're not dependent on deepstream's internal objects ("artificial" ones)

src/nvmsgbroker_pipeline.py
lyuzinmaxim: It was a very long debugging together with chatgpt...

To me your code also looks fine, to be honest, I don't know what (and why) it didn't work.

At the end, I just followed documentation and reused the same default `NvDsPersonObject`, and populated its fields with our data (e.g., `person.gender = "plant", apparel = vevent.prediction.class_name` etc)

PR #163, opened 2025-11-18
Debug/basic frame comparison

### LINKED ISSUE #156: Debug frame comparison
We want to understand the reason why frame comparison thing is not working in the deepstream pipeline as expected.

For doing that, we need to localize the issue:
- a simple opencv code that reads the video (any of 4 videos `assets/videos/frame_change_videos`), with explicit logging which frame was "skipped", and which was "dropped"
- frame comparison pipeline (`src/frames_skipping_pipeline.py`) provides exactly the same logging. 

We want to compare the results (skipped vs dropped frame-by-frame) from those 2.

DoD:
- opencv code (case 1) is in the repository
- Comparison of opencv and deepstream is shared in slack for each of 4 videos

Sarah5567:
## Summary

This PR adds a minimal OpenCV-based frame comparison script used to validate DeepStream frame skipping behavior. The goal was to localize differences between the CPU (OpenCV) and GPU (DeepStream) implementations by comparing which frames were **processed** and which were **skipped**.

## What was done

- Added a simple OpenCV frame comparison script that logs per-frame decisions (process/skip)
- Ran comparisons on all 4 test videos and matched the results against the DeepStream pipeline output
- Identified that the code logic is correct, and the mismatch originated from a **specific corrupted video**, not from the DeepStream implementation
- Threshold tuning is still required, but will be finalized only when the final production video is available

closes #156


src/frame_comparison/cv_frame_skip_debug.py
```diff
@@ -0,0 +1,49 @@
...
+def run_on_directory(directory: str) -> None:
+    """Run analysis on all supported video files in a directory."""
+    for filename in os.listdir(directory):
+        if filename.lower().endswith(VIDEO_EXTENSIONS):
+            video_path = os.path.join(directory, filename)
+            analyze_video_skip_decisions(video_path)
+
+
+if __name__ == "__main__":
+    run_on_directory(VIDEO_DIR)
```
r83575, line 49: Maybe you can think of a more appropriate name for the file.

PR #168, opened 2025-11-20
Align C++ frame skipping probe logic with Python implementation

### LINKED ISSUE #165: Change frame comparison C++ implementation to be in sync with python one
In frame comparison GPU implementation, we have it different from python-based one: e.g., we do not run resizing. That's totally fine.

In this ticket, we want to keep the FrameChangeDetector class as is, but update the probe/helpers logic to be the same as in python.

Reference: https://github.com/KamaTechOrg/AgStream/pull/158

DoD:
- logic in probe/helpers/... is updated to be consistent with python implementation (no singleton, providing class instances as a parameter, etc)

r83575:
This PR updates the GPU frame comparison probe logic in C++ to match the logic used in the Python implementation.

Changes include:
- Removed singleton usage
- Cleaned up probe logic and ensured same frame filtering behavior as in Python

Closes #165


src/deepstream/pipelines/gpu_frames_skipping_pipeline.py
```diff
@@ -22,7 +22,8 @@
 def jetson_frame_skip_probe(pad: Gst.Pad, info: Gst.PadProbeInfo, u_data: Optional[Any]) -> Gst.PadProbeReturn:
     buffer_ptr: int = hash(info.get_buffer())
     batch_id: int = 0
-    should_process: bool = frame_skipping_probe.frame_skipping_probe(buffer_ptr, batch_id)
+    detector: GPUFrameChangeDetector = u_data
```
Sarah5567: Consider declaring `detector: GPUFrameChangeDetector` directly in the function parameters instead of assigning it later.

src/deepstream/pipelines/gpu_frames_skipping_pipeline.py
```diff
@@ -20,10 +19,10 @@
 
 stats: Dict[str, int] = {"total": 0, "skipped": 0, "processed": 0}
 
-def jetson_frame_skip_probe(pad: Gst.Pad, info: Gst.PadProbeInfo, u_data: Optional[Any]) -> Gst.PadProbeReturn:
+def jetson_frame_skip_probe(pad: Gst.Pad, info: Gst.PadProbeInfo, detector: GPUFrameChangeDetector) -> Gst.PadProbeReturn:
```
lyuzinmaxim: I would not have the platform name in function name - we don't gain any information from "jetson" (it can also be a cpu-based implementation, for example). I'd use `gpu_frame_skip_probe` or smth similar.

src/deepstream/pipelines/gpu_frames_skipping_pipeline.py
lyuzinmaxim, line 1: same as [here](https://github.com/KamaTechOrg/AgStream/pull/168/files#r2554122750)

Could you please remove the jetson from function names, e.g. `build_jetson_pipeline` -> `build_pipeline`?

Also I think your branch is not up-to-date with main: [runner](https://github.com/KamaTechOrg/AgStream/blob/a4588523707898e586e4d1820ce065abf5f83d37/src/deepstream/helpers/pipeline_runner.py) is not used

PR #170, opened 2025-12-01
update the threshold values

r83575:
I checked the 4 videos for my [PR](https://github.com/KamaTechOrg/AgStream/pull/168) - Align C++ frame skipping probe logic with Python implementation #168.

And I notice that the thresholds in the GPU version are not tuned well,
When they were tuned it's was on 4 colorful videos and probably this Is the reason.
This PR fixes this issue.


src/deepstream/probes/frame_comparison/gpu/frame_change_detector.cpp
```diff
@@ -5,6 +5,11 @@
 #include <cmath>
 #include <vector>
 
+const double MSE_THRESH = 20.0;
+const double SSIM_THRESH = 0.998;
+const double FLOW_THRESH = 0.5;
+const double OPTICAL_FLOW_ACTIVE_THRESH = 0.5;
```
Sarah5567, line 11: Great that you wrote them as constants, but please remove the initialization from the constructor or use these constants as the initial values there.

lyuzinmaxim (approved):

PR #171, opened 2025-12-07
Final gpu pipeline

### LINKED ISSUE #140: Create a final pipeline (frame comparison GPU and background removal GPU)

Sarah5567:
## Summary
Added `deepstream_pipeline_gpu.py`, combining the GPU frame-skipping probe, GPU background-removal module, and modular nvmsgbroker metadata probe into a single unified pipeline.

## Result
An end-to-end DeepStream pipeline that integrates all required components and runs successfully.

closes #140


src/deepstream/probes/frame_comparison/gpu/frame_change_detector.cpp
```diff
@@ -5,8 +5,8 @@
 #include <cmath>
 #include <vector>
 
-const double MSE_THRESH = 20.0;
-const double SSIM_THRESH = 0.998;
+const double MSE_THRESH = 99.0;
```
lyuzinmaxim, line 8: why do we need to change it?

